PR #24, opened 2015-06-10
update lint and config

dougfelt:
- if lint detects path intersection, lint now outputs a bit more information.
  (For intersection, points on the pieces of the contours that intersect.)
- lint config rules exclude some tests for CJK fonts based on the v1.003 drop
  and emails with Adobe.

### DIFF
--- a/nototools/lint_config.txt
+++ b/nototools/lint_config.txt
@@ -2,8 +2,8 @@
 disable head/os2/unicoderange
 
 vendor Monotype
-disable paths # no extrema
 disable trademark # we can swat this ourselves
+disable extrema
 
 filename like Avestan
 disable gpos/missing
@@ -11,3 +11,23 @@ disable gpos/missing
 filename like Bengali
 enable script_required except cp 951-952
 
+condition
+vendor Adobe
+
+# Adobe's position is that these (mathematical) characters are intended to be used with CJK
+# in an LTR environment.  For true bidi text (that would naturally trigger bidi) these
+# characters should come from a font designed for bidi.  So there is no need in this font
+# for the mirrored glyphs.
+disable bidi
+
+# The v1.003 fonts align the windows ascent/descent with the hhea values, but not the
+# typographic ascender
+version == 1.003
+disable head/os2/ascender, head/os2/descender
+
+# Adobe says "None of the zero-width glyphs are involved with attachment"
+disable gdef/classdef/not_present
+
+# Adobe says that most of the glyphs are indeed reachable, so perhaps the linter/ttx is
+# at fault?
+disable reachable
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -286,6 +286,7 @@ def to_float_tuples(curve):
 
 def curves_intersect(contour_list):
     """Takes a list of contours and tells if any two curves in them intersect.
+    Returns a message string with an error, or an empty string if ok.
     """
     all_contours = []
     for contour in contour_list:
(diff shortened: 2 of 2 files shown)


nototools/lint_config.txt
```diff
@@ -11,3 +11,23 @@ disable gpos/missing
...
+# typographic ascender
+version == 1.003
+disable head/os2/ascender, head/os2/descender
+
+# Adobe says "None of the zero-width glyphs are involved with attachment"
+disable gdef/classdef/not_present
+
+# Adobe says that most of the glyphs are indeed reachable, so perhaps the linter/ttx is
+# at fault?
+disable reachable
```
roozbehp: I don't like the reachability conclusion at all. What is one such example that the glyph is reachable but ttx is finding it unreachable? If there is such example, where is the linter/ttx bug for that?

nototools/lint_config.txt
```diff
@@ -11,3 +11,23 @@ disable gpos/missing
...
+# characters should come from a font designed for bidi.  So there is no need in this font
+# for the mirrored glyphs.
+disable bidi
+
+# The v1.003 fonts align the windows ascent/descent with the hhea values, but not the
+# typographic ascender
+version == 1.003
+disable head/os2/ascender, head/os2/descender
+
+# Adobe says "None of the zero-width glyphs are involved with attachment"
```
roozbehp: @dougfelt, what is the current list of these glyphs for the Noto CJK fonts?

dougfelt: Run the lint tool on the CJK fonts, which were pushed yesterday.  There are hundreds.

For example. NotoSansCJKsc-Thin.otf reported this:
"The following 919 glyphs are unreachable in the font: cid01357,
> cid01676, cid04122, cid06950, cid13116, cid13678, cid61783, cid61784,
> cid61785, cid61788, cid61789, cid61790, cid61791, cid61793, cid61794,
...

Adobe replied:

"Except for a very small number of exceptions (whose glyphs are likely to be exorcized in the Version 2.000 update), all glyphs are reachable by some means. Sometimes, it depends on the font, due to default language. CID+1676, for example, is reachable via the 'hist' GSUB feature. Sometimes, more than one GSUB feature needs to be used, such as 'ccmp' then 'vert'."

We need to investigate this.  There are other issues with the unreachable glyph implementation as well.  It seems reasonable at the moment to disable the test while we investigate and figure out why we're seeing such a difference from what Adobe sees with their tools.

roozbehp: Ok. Regarding the unreachable glyphs, please file a bug against linter mentioning this and that we should re-enable the test after we fix it.

What about the zero-width glyphs not having classdef? What are they?

dougfelt: @roozbehp:

> "There is no GlyphClassDef subtable of GDEF table in the font, while
> there are non-spacing combining characters: 302A..302D, 3099..309A."

I suspect that this is okay, because our tools didn't find a need to create this particular table. None of the zero-width glyphs are involved with attachment, and that is likely why this table is not needed.

dougfelt: The comment above "I suspect that..." comes from Adobe, if it wasn't clear.

dougfelt: Filed bug about the reachability problem in nototools.

roozbehp:
@behdad, what happens with CJK combining marks in HarfBuzz when there's no GDEF/classdef? These are the characters I'm concerned about: U+302A..U+302D, U+3099, and U+309A.

behdad:
> @behdad, what happens with CJK combining marks in HarfBuzz when there's no GDEF/classdef? These are the characters I'm concerned about: U+302A..U+302D, U+3099, and U+309A.

I believe they will be marked as marks (pun intended!) based on Unicode category.

PR #28, opened 2015-06-11
add tools for udhr samples and ohchr attributions

dougfelt:
This code manages samples based on article 1 of the universal declaration of human rights.  We have been using  some translations obtained by hand, this bases the samples on data obtained from unicode.org/udhr and attribution data from ohchr.org and uses tooling to put it together.

It's not guaranteed that the new sample data is better than the old-- I've noticed and fixed some typos-- but this lets us base our versions on a known source.

The attribution data is used for notices ohchr/un have requested we show when using these translations.  Unfortunately different translations require different notices.

### DIFF
--- a/nototools/extract_ohchr_attributions.py
+++ b/nototools/extract_ohchr_attributions.py
@@ -69,8 +69,11 @@
 
 
 import argparse
+import codecs
 import HTMLParser as html
 import re
+
+from nototools import tool_utils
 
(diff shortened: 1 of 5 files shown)


dougfelt:
Unless someone objects today, I'm going to merge the pull request so the new website generation code I'm working on can more easily make use of it.

behdad:
Go for it.

PR #34, opened 2015-06-19
Fix setup.py

khaledhosny:
add_emoji_gsub.py and map_pua_emoji.py were removed in
fb5b819093b9a6500262653cc8b1e1188ad0a663.


behdad:
Thanks!

PR #35, opened 2015-06-19
New website generation code

dougfelt:
This is primarily a rewrite of the generate_website_data code.  The website structure and data organization has changed, which was the primary motivation. Some other changes were made along with this:
- sample texts sourced from the UN have attributions
- additional script samples were added so that no font lacks a sample
- cldr-specific code was moved into its own file, and the generate_samples_from_exemplars script was updated to use it


dougfelt:
No one seems to care, so in it goes.

PR #56, opened 2015-07-15
Remove redundant '?' in regex.

dougfelt:
Python versions before 2.7.6 complain about '(.*)?' since
the '?' is redundant.  Regex compilation should strip this, but didn't
before then, and reported 'sre_constants.error: nothing to repeat'.
Clients should use more recent versions of the python 2.7 stream as
there might be more issues than this one; I fixed this mostly as reference.


behdad:
Such a weird warning / error!

PR #74, opened 2015-08-27
Grab bag of mostly changes

dougfelt:
See individual commits for most things.

The primary driver was to refactor lint so that it shared the noto_fonts filename parsing code with generate_website_2_data.  I also wanted to make it easier to run lint on large batches of files, so there are some feature changes related to this.

There are also fixes to some of the data used in website generation, to address some reported issues.


jamesgk:
LGTM

PR #77, opened 2015-09-01
Support Emoji in web data, fix Ekajuk script

dougfelt:


nototools/generate_website_2_data.py
```diff
@@ -644,7 +644,9 @@ def build_data_json(self, family_id_to_lang_scr_to_sample_key,
       else:
         num_fonts = len(family.hinted_members or family.unhinted_members)
       family_obj['fonts'] = num_fonts
-      family_obj['langs'] = len(family_id_to_lang_scr_to_sample_key[k])
+      # only displayed langs -- see build_family_json lang_scrs
+      lang_scrs_map = family_id_to_lang_scr_to_sample_key[k]
+      family_obj['langs'] = sum([1 for l in lang_scrs_map if not l.startswith('und-')])
```
jamesgk: Interesting to note that you don't need the brackets in `sum` here -- it can be passed a generator. Not sure that's worth another commit, though.

jamesgk:
LGTM

PR #80, opened 2015-09-05
Spreadsheet

dougfelt:
Modified the coverage tool to provide the options of dumping the charmap as a range list, and as a text file, to make it easier to generate some sample text to view.

### DIFF
--- a/nototools/generate_lang_font_table.py
+++ b/nototools/generate_lang_font_table.py
@@ -1,0 +1,89 @@
+#!/usr/bin/python
+#
+# Copyright 2015 Google Inc. All rights reserved.
+#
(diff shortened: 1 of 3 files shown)


dougfelt:
@jungshik, can you look at a612fc6 (Generate table mapping from language to font name).  This is what I used to generate the table I sent in the email.  I still need to update it to handle the exceptions you pointed out in your email (preferred fonts for Arabic and Urdu).  It seems better to me to encode these suggestions in the tool.

@jamesgk I put you as reviewer so you can check my sloppy python.

I really wanted these to be in separate pull requests, but I had insufficient git-fu to disentangle them.

jamesgk:
LGTM

PR #83, opened 2015-09-10
Swat license and other changes

dougfelt:
Font_data changed to allow writing a new name record, which swat needs.
Includes some unrelated lint refactoring, more in a separate cl.
Also fixes a bug in compare_summary.

### DIFF
--- a/nototools/swat_license.py
+++ b/nototools/swat_license.py
@@ -29,9 +29,29 @@ from nototools import noto_fonts
 from fontTools import ttLib
 from fontTools import misc
 
+_COPYRIGHT_ID = 0
 _VERSION_ID = 5
+_TRADEMARK_ID = 7
+_MANUFACTURER_ID = 8
+_DESIGNER_ID = 9
+_DESCRIPTION_ID = 10
+_VENDOR_URL_ID = 11
+_DESIGNER_URL_ID = 12
 _LICENSE_ID = 13
 _LICENSE_URL_ID = 14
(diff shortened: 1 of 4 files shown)


nototools/swat_license.py
```diff
@@ -0,0 +1,319 @@
...
+  'Arimo': 'a',
+  'Cousine':'b',
+  'Tinos': 'c',
+  'Noto': 'd',
+}
+
+_CHANGES = {}
+
+def _swat_fonts(dst_root, dry_run):
+  def family_key(family):
```
jamesgk: Is this and the `_FAMILY_KEYS` dictionary unused?

nototools/swat_license.py
```diff
@@ -0,0 +1,319 @@
...
+}
+
+_FAMILY_KEYS = {
+  'Arimo': 'a',
+  'Cousine':'b',
+  'Tinos': 'c',
+  'Noto': 'd',
+}
+
+_CHANGES = {}
```
jamesgk: I think this name should not be all caps since the dictionary is going to change.

nototools/swat_license.py
dougfelt: Thanks for noticing this!

nototools/swat_license.py
dougfelt: OK

jamesgk:
LGTM

PR #84, opened 2015-09-11
Clean up lint of cjk

dougfelt:
This splits out computation of the expected names for cjk/non-cjk into their own smaller methods, so that we can have one main method to check the name table rather than one for cjk and one for non-cjk, which was formerly the case.


jamesgk:
LGTM

PR #90, opened 2015-09-25
Move some data extraction methods from Roboto

jamesgk:
This is one small initial step in moving testing stuff from Roboto to Noto. Definitely open to feedback wrt whether I'm putting stuff in the right place, or if stuff is general enough.


jamesgk:
Right, I hadn't seen noto_fonts.py. So I'll submit another pull request altering that file instead.

PR #92, opened 2015-09-29
Update lint, web gen for OFL and emoji font location changes

dougfelt:


nototools/generate_website_2_data.py
```diff
@@ -246,8 +246,7 @@ def get_family_id_to_regions(family_id_to_lang_scr_to_sample_key):
 
   family_id_to_regions = collections.defaultdict(set)
   warnings = set()
-  for (family_id, lang_scr_to_sample_key in
-       family_id_to_lang_scr_to_sample_key.iteritems()):
+  for family_id, lang_scr_to_sample_key in family_id_to_lang_scr_to_sample_key.iteritems():
```
jamesgk: Accidental change? The line is now over 80 characters.

jamesgk:
LGTM

nototools/generate_website_2_data.py
dougfelt: When I tried to run this with a wrapped line, it failed.  Thought I had run it with the wrapped line, but maybe I hadn't.  Anyway, I had to unwrap it to make it work.

Should the wrapped version work?

roozbehp: Try unpacking the tuple inside the loop?

dougfelt: That is frighteningly obvious, isn't it... Not my day.

On Tue, Sep 29, 2015 at 2:19 PM, Roozbeh Pournader <notifications@github.com

PR #96, opened 2015-10-02
Fixes related to web site generation

dougfelt:
Main motivation is to fix some issues around language / script presentation on the web site, prompted by #485.  This slightly changes the language/script/region association of some Chinese dialects, and also restores Okinawan as a language supported by Noto Sans CJK JP.

The rest is a (relatively failed) attempt to clean up how CLDR data is used to drive this.


jamesgk:
LGTM

PR #98, opened 2015-10-06
Add unit tests to nototools/unittests

jamesgk:
These new files are almost identical to those in https://github.com/google/roboto/tree/master/scripts (with some changes to import statements), so this shouldn't need too much review despite its size.


dougfelt:
LGTM

PR #100, opened 2015-10-14
Create tool to generate variation selector cmap.

dougfelt:
Adds data and tool, and updates setup.py so noto-emoji will be able to use it.


third_party/ucd/emoji-data.txt
```diff
@@ -0,0 +1,1315 @@
+# Emoji Data for UTR #51
```
roozbehp: You need to update the README.third_party file to point to location and version of this file.

nototools/add_vs_cmap.py
```diff
@@ -0,0 +1,124 @@
...
+VS_TEXT = 0xFE0E
+VS_EMOJI = 0xFE0F
+
+def modify_font(font_name, font, presentation, emoji_variants):
+  cmap_table = font_data.get_cmap(font)
+  emoji = set(cmap_table.keys()) & emoji_variants
+  if not emoji:
+    print 'no emoji match those in %s' % font_name
+    return
+  uvs = VS_EMOJI if presentation == 'emoji' else VS__TEXT
```
roozbehp: Why two underscores? (pylint should be able to catch these.)

dougfelt: because that's a typo...

dougfelt:
Thanks Roozbeh.

FYI, the protocol James and I have been following is the reviewer LGTMs the
pull request, and then the person who initiated it (who generally has write
permission in all these cases) merges it.

On Wed, Oct 14, 2015 at 4:38 PM, Roozbeh Pournader <notifications@github.com

roozbehp:
Sure. OK.

PR #102, opened 2015-10-16
Change coverage.py tool to show private use characters.

dougfelt:
Generating text was filtering out all characters with general category
C, which excluded PUA characters. This lets coverage.py report on them.

Also adds option to customize the separator put between characters
when generating a text file.


jamesgk:
LGTM

PR #105, opened 2015-10-21
Add a tool for generating text samples based on glyphs in a font.

dougfelt:
Some bug reports refer to names of characters in a particular script, but
don't provide the unicode values of these characters.  This tool lets
you interactively enter (portions of) names of characters and match these
names against the unicode names of characters available in the font to
build up a text string, which you can then write to a file.  This makes
it easier to build example text to view in that font.

### DIFF
--- a/nototools/sample_with_font.py
+++ b/nototools/sample_with_font.py
@@ -137,9 +137,9 @@ def _get_char_names(charset):
       except:
         name = None
       if not name or name == '<control>':
-        name = 'uni_%04x' % cp
+        name = '%04x' % cp
       else:
-        name = name.lower()
+        name = '%04x %s' % (cp, name.lower())
       name_map[name] = cp
 
   return name_map
--- a/nototools/sample_with_font.py
+++ b/nototools/sample_with_font.py
@@ -122,7 +122,7 @@ def _build_text(name_map, initial_text=''):
         continue
 
       text += unichr(name_map[matches[n]])
-      select_multiple = false
+      select_multiple = False
 
   print 'done.'
   return text
--- a/nototools/sample_with_font.py
+++ b/nototools/sample_with_font.py
@@ -111,7 +111,7 @@ def _build_text(name_map, initial_text=''):
         try:
           n = int(line)
           break
-        except:
+        except ValueError:
           continue
 
       if not select_multiple: # q
--- a/nototools/sample_with_font.py
+++ b/nototools/sample_with_font.py
@@ -1,0 +1,160 @@
+#!/usr/bin/python
+#
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


nototools/sample_with_font.py
```diff
@@ -0,0 +1,160 @@
...
+      if line == 'q':
+        continue
+      try:
+        n = int(line)
+        if n < 0 or n >= len(matches):
+          print '%d out of range' % n
+          continue
+        text += unichr(name_map[matches[n]])
+        break
+      except:
```
jamesgk: What are we expecting to except here? Can we do without this entirely?

nototools/sample_with_font.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if not new_matches:
+      new_matches = matches
+
+    while True:
+      print 'multiple matches:\n  ' + '\n  '.join(
+          '[%2d] %s' % (i, n) for i, n in enumerate(matches))
+      line = raw_input('0-%d or q > ' % (len(matches) - 1))
+      if line == 'q':
+        continue
```
jamesgk: Should this be break?

nototools/sample_with_font.py
```diff
@@ -0,0 +1,160 @@
...
+
+  print 'done.'
+  return text
+
+
+def _get_char_names(charset):
+  name_map = {}
+  if charset:
+    for cp in charset:
+      name = unicode_data.name(cp)
```
jamesgk: Do we want to accept/pass on PUA characters? If so, we can except a ValueError on this line.

nototools/sample_with_font.py
dougfelt: the line input might not parse as an int.  could be random text.

nototools/sample_with_font.py
dougfelt: Yes, hmmm must have moved this and not retested it.

nototools/sample_with_font.py
dougfelt: Thanks, yes need to handle that.

nototools/sample_with_font.py
jamesgk: Ah, right. I think it would be better to only except a ValueError at `n = int(line)`.

nototools/sample_with_font.py
```diff
@@ -0,0 +1,184 @@
...
+
+      if not select_multiple: # q
+        break
+
+      if n < 0 or n >= len(matches):
+        print '%d out of range' % n
+        continue
+
+      text += unichr(name_map[matches[n]])
+      select_multiple = false
```
jamesgk: False should be capitalized

dougfelt: Man, I'm just batting 1000 on this aren't I.

jamesgk:
LGTM. This is a nice tool!

PR #107, opened 2015-10-28
Add variant selector lint tests.

dougfelt:
This requires the hb-shape tool from Harfbuzz.  Python expects to be
able to call this tool using subprocess with the name 'hb-shape' when
running the GSUB-based tests.  So 'hb-shape' must be in a directory
on your PATH.

This does several things:
- adds functions to unicode_data.py that provide variant information
  on the non-emoji variants in StandardizedVariants.txt.  A previous
  submit handled only the emoji case.  Emoji variants are rather
  different from other variants, so it seemed reasonable to handle them
  separately.
- add lint tests to check that if a font contains characters that have
  standard variants, that it has a format 14 cmap and that the behavior
  corresponds to the behavior expected in the cmap (in particular, if
  a variant sequence corresponds to a CJK compatibility character, the
  character is in the cmap and the glyph for the sequence is the same as
  that for the character).
- add GSUB tests to check that the GSUB table will generate single glyphs
  for the standard variant sequences in the font.  This assumes the font
  generates single glyphs in this case, which is true for the two Noto
  fonts that currently do this, Mongolian and PhagsPa.  This uses
  harfbuzz to generate the glyphs.
- updates lint config to disable the GSUB based test by default, and
  enable it (and disable the cmap based test) for Mongolian, PhagsPa,
  and Historic.
- updates lint config to disable cmap tests for Adobe's CJK fonts. They
  carefully pick a subset of the standard variation sequences to support.

### DIFF
--- a/nototools/lint_config.py
+++ b/nototools/lint_config.py
@@ -348,6 +348,7 @@ class TestSpec(object):
     private_use
     non_characters
     disallowed_ascii
+    variants
   head -- head table tests
     hhea
       ascent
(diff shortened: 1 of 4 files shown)


jamesgk:
LGTM

PR #109, opened 2015-11-07
add arg to create_image to set output filename

dougfelt:


nototools/create_image.py
```diff
@@ -20,9 +20,10 @@
 __author__ = 'roozbeh@google.com (Roozbeh Pournader)'
 
 import argparse
+import codecs
```
jamesgk: Is this still used?

nototools/create_image.py
```diff
@@ -198,29 +208,34 @@ def _get_style(style_name):
...
+      name_strs = [font.replace(' ', '')]
+      name_strs.extend(['%x' % ord(cp) for cp in text])
+      if weight_name:
+        name_strs.append(weight_name)
+      if style_name:
+        name_strs.append(style_name)
+      name_strs.append(str(font_size))
+      if lang:
+        name_strs.append(lang)
+        file_name = '_'.join(name_strs) + '.' + ext
```
jamesgk: Should this line be un-indented?

nototools/create_image.py
dougfelt: I removed the import from the other location.  My java bias leads me to prefer putting the imports at the top of the file, even it if means we import some things we don't use unless we make a particular call.

nototools/create_image.py
dougfelt: Sure should, thanks.

nototools/create_image.py
jamesgk: Oh, I don't think it was actually removed in this commit.

dougfelt:
sorry, replied before I'd pushed.  now the push is up.

jamesgk:
LGTM

PR #111, opened 2015-11-09
make unicoderange more accurate

dougfelt:
Two small related changes, prompted by running autofix on Javanese.
- autofix_for_release now reports both the old and new unicoderange info
  when it changes it.  Formerly it only reported the new info, so it
  was not possible to tell what changed.
- unicode_data now only sets the bit for a range when characters actually
  fall into the range.  formerly it was setting it when a character
  was <= the endpoint of the current range.  This presumed no gaps in
  the ranges, but there are gaps.


jamesgk:
LGTM

PR #117, opened 2015-11-19
Add non-udhr classical chinese text sample

dougfelt:
For the fun of it, let's actually use a classical Chinese text sample.


jamesgk:
LGTM

PR #118, opened 2015-11-20
disable pua and gpos checks for emoji

dougfelt:
For the fun of it, let's actually use a classical Chinese text sample.

### DIFF
--- a/nototools/lint_config.txt
+++ b/nototools/lint_config.txt
@@ -30,6 +30,8 @@ disable bounds/font/ymax
 
 filename like Emoji
 disable filename/name
+disable cmap/private_use
+disable complex/gpos/missing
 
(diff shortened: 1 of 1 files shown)


roozbehp:
I would rather change this to check for existence of the specific PUA emojis we need for backward compatibility and making sure there's no other PUA characters in the font.

dougfelt:
So, that set is exactly fe4e5-fe4ee fe82c fe82e-fe837?

NotoEmoji (which Android doesn't use, right?) has the keycap emoji at e023
(#) and e030-e039.  Should they not be there at all?  Or should they be
mapped to fe82c fe82e-fe827?

On Thu, Nov 19, 2015 at 5:45 PM, Roozbeh Pournader <notifications@github.com

PR #122, opened 2015-11-26
Generalize glyph area unit test

jamesgk:
I left some Roboto-specific stuff in here before; this makes the test
more general. It now runs on both UFOs and TTFonts, with any number of
instances, and the area pen can now handle glyphs with cubic or
quadratic curves


dougfelt:
lgtm

PR #124, opened 2015-12-02
Update website generation tools for noto color emoji

dougfelt:
The Nov. 30 changes are a bit separate from this, there's another pull request for them that should go in before this one.


jamesgk:
LGTM

PR #125, opened 2015-12-02
add tool to generate emoji name data

dougfelt:
.This is to support the website, we want to generate a table of names for those emoji images that we want to display.

### DIFF
--- a/nototools/generate_emoji_name_data.py
+++ b/nototools/generate_emoji_name_data.py
@@ -14,12 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Generate name data for emoji resources."""
+"""Generate name data for emoji resources. Currently in json format."""
 
 import argparse
 import collections
 import glob
-import json
 import os
 from os import path
 from nototools import unicode_data
--- a/nototools/generate_emoji_name_data.py
+++ b/nototools/generate_emoji_name_data.py
@@ -63,13 +63,18 @@ def flag_sequence_name(cps):
   return ''.join(unichr(cp - REGIONAL_INDICATOR_A + ord('A')) for cp in cps)
 
 
+_NAME_FIXES = {
+    'Oclock': "O'Clock",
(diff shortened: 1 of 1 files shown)


nototools/generate_emoji_name_data.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def unicode_name(cp):
+  name = unicode_data.name(cp).title()
+  name = name.replace('Oclock', "O'Clock")
+  name = name.replace('Mans', "Man's")
+  name = name.replace('Womans', "Woman's")
+  name = name.replace('Mens', "Men's")
+  name = name.replace('Womens', "Women's")
+  return name
```
roozbehp: I would change the data to a dictionary and loop over it.

nototools/generate_emoji_name_data.py
```diff
@@ -0,0 +1,142 @@
...
+    if len(cps) == 1:
+      name = unicode_name(cps[0])
+    elif is_flag_sequence(cps):
+      name = flag_sequence_name(cps)
+    else:
+      name = ''
+    output[tuple(cps)] = (fname, sequence, name)
+
+  with open(outfile, 'w') as f:
+    f.write('[\n')
```
jamesgk: Any reason we're not using the Python json library here?

dougfelt: So I don't have to build objects from the data just so the json library can decode them to generate the output?  The data is really simple so it didn't seem necessary.

nototools/generate_emoji_name_data.py
dougfelt: OK, changed that.

nototools/generate_emoji_name_data.py
jamesgk: Yea I guess it's simple either way. I would always tend towards using a library for peace of mind. But if we don't use it here we have an unused `import json` statement above.

jamesgk:
LGTM

PR #126, opened 2015-12-02
Javanese lint config changes

dougfelt:


jamesgk:
LGTM though I don't know the rationale for these changes. I guess my only feedback is that more comments might be useful.

PR #128, opened 2015-12-04
Fix lint assumption that all non-cjk fonts are from Monotype.

dougfelt:
Also re-enables name table trademark and vendor_url checks.


jamesgk:
LGTM

PR #133, opened 2015-12-11
[unittests] Fix up name table test

jamesgk:
With this change the test no longer expects "Regular" in a font name
unless that is the complete font name. It also passes more flexible
parameters to the function which generates unique IDs (defined by
subclasses).


dougfelt:
LGTM

PR #140, opened 2015-12-17
Tweak noto_lint for use with development, monospace fonts.

dougfelt:
Some development drops omit some data we were expecting, so
be a bit more robust.

This also changes the advances/whitespace check to expect whitespace
in monospace fonts to have the same default advance (except for ZWS),
instead of the advances implied by their definitions.

### DIFF
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -64,7 +64,7 @@ SIL_LICENSE = ("This Font Software is licensed under the SIL Open Font License, 
 
 SIL_LICENSE_URL = "http://scripts.sil.org/OFL"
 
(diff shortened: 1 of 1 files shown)


nototools/noto_lint.py
```diff
@@ -1707,7 +1710,12 @@ def check_for_bidi_pairs(cmap):
 
         rtlm = {}
         if "GSUB" in font:
-            feature_record = font["GSUB"].table.FeatureList.FeatureRecord
+            try:
+              feature_record = font["GSUB"].table.FeatureList.FeatureRecord
+            except:
```
jamesgk: Maybe just except an AttributeError here

jamesgk:
LGTM

dougfelt:
Amended the commit with the suggested change.

PR #144, opened 2015-12-18
Update UDHR sources and samples.

dougfelt:
Renames emoji sample file after the new script code.  Also fixes a few issues with the tooling.


jamesgk:
LGTM

jamesgk:
Huh, swb-Arab_udhr.txt was added here, but had been previously renamed to swb-Latn_udhr.txt: https://github.com/googlei18n/nototools/commits/master/sample_texts/swb-Arab_udhr.txt. Do we need to rename it again? Fix something in the tools that generate sample texts?

dougfelt:
I need to fix the tools to scan the text and minimally verify that it is in the script we expect it to be in.

PR #149, opened 2016-01-08
Add more scripts/chars to extra_chars_needed, including LGC chars.

dougfelt:


dougfelt:
Sorry, got git into an odd state and don't have the git-fu to clean it up.

dougfelt:
Thanks to @moyogo for showing me how I can use branch renaming to clean this up.

dougfelt:
@roozbeh ping?

dougfelt:
@ jungshik, reassigned to you since roozbeh is too busy.

So far this is just a data change.  The main use of this data in nototools is by lint, it generates a list of unicode characters that need to be in the font cmap, based on the script(s) of the font. By default lint assumes all characters with that primary script or with that script in their script extensions belong in the font (except for Nastaliq/Aran, which is special-cased).  This misses common characters, as well as cases where characters from other blocks are used with a script but script extensions does not include it.  When common punctuation is used with a script I want to list it if it might at all interact typographically (e.g. kerning, positioning, or stylistic changes), this should help ensure more focus on these characters to see if they do interact.  To determine what is needed I'm looking at the unicode L2 history of the script, our existing fonts, and some other data of various degrees of reliability like cldr exemplars and web pages, it's kind of ad-hoc.

dougfelt:
About Coptic Epact in Coptic, @roozbehp explicitly didn't want this.  He thinks they should go into a symbols font.  Some comments he made in an email thread when I was asking about Coptic EPACT and Arabic Mathematical characters:

"Coptic Epact numbers and Arabic Mathematical characters don't need different weights. So they can both be just in a single-weight symbols font."

"All in all, the Arabic fonts themselves don't need either the Arabic Mathematical characters or the Coptic Epact numbers."

"Please don't put them in Noto Sans Coptic. That would make them very hard to find. IIRC, these numbers don't have much of a relation with the Coptic script, but with the writing of the Copts in the Arabic script."

jungshik:
Thanks a lot for the reply.  LGTM, then.

PR #152, opened 2016-01-15
Add a "fontdiff" package for comparing fonts

jamesgk:
This might be generally useful for comparing a font against another
font which you know to be correct. Currently tools here compare glyph
shapes and positioning (GPOS) features.

The tools try to put largest diffs first, so even if there are many
diffs you can just look at the first few, which should be the most
significant. For glyph shapes, multiple pairs of fonts can be compared
at once and the diffs from all pairs sorted together.

For positioning features, ttxn output is compared rather than actual
font binaries. ttxn is a tool bundled in the AFDKO, see description:
http://www.adobe.com/devnet/opentype/afdko/topic_overview.html
Effectively, de-compiled feature syntax is being compared rather than
the GPOS tables.

Sorry for the large PR! I should have pushed this earlier....

cc @davelab6

### DIFF
--- a/nototools/fontdiff/__main__.py
+++ b/nototools/fontdiff/__main__.py
@@ -1,3 +1,28 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


davelab6:
Awesome!! Thanks for making this available to the community :D

dougfelt:
LGTM modulo comments.  Sorry for forgetting about this.

davelab6:
Thanks guys!

PR #156, opened 2016-01-22
Update file name parsing for phase 3, and refactor lint

dougfelt:
Adds noto_names.py to hold methods related to noto font naming and name table entries, and refactors noto_lint to use it.


dougfelt:
@jamesgk sorry for the large commits, nothing too hairy is going on though.

noto_lint now depends on a data file that is generated by the noto_names tool, but I didn't check that in or add a warning for it in lint.  Still have to figure out whether the phase should affect this data directly, or just be applied in lint.

nototools/noto_lint.py
```diff
@@ -736,6 +603,47 @@ def _check_unused_names():
           warn('name/unused', 'Name', 'Name table has record #%d: "%s"' %
                (i, names[i]), is_error=False)
 
+    def _noto_font_from_font_props(font_props):
+      fields = """
+          filepath,family,style,script,variant,width,weight,slope,fmt,
+          manufacturer,license_type,is_hinted,is_mono,is_UI,is_display,
+          is_cjk,subset
```
jamesgk: It seems like this same attribute list is replicated in a few places (here and in the NotoFont and FontProps definitions), could it be defined once somewhere?

nototools/noto_names.py
```diff
@@ -0,0 +1,711 @@
...
+    'LGC': None,
+    'Zsye': None,
+}
+
+
+# copied from noto_lint, we should have a better place for it.
+def preferred_script_name(script_key):
+  try:
+    return unicode_data.human_readable_script_name(script_key)
+  except:
```
jamesgk: Could this catch a more specific exception? There's another `except:` later in the file also.

jamesgk:
Just minor comments, LGTM

jamesgk:
Oh, something else -- a unit test needs to be updated due to changes to noto_fonts in ede5ab9c790f39b9a146560a311a0773feb1e2f7:

```diff
--- a/nototools/unittests/font_tests.py
+++ b/nototools/unittests/font_tests.py
@@ -178,7 +178,7 @@ class TestNames(FontTest):
         family_names = '%s|%s' % (condensed_name_nosp, name_nosp)
 
         filename_match = noto_fonts.match_filename(filename, family_names)
-        family, _, _, _, _, weight, slope, _ = filename_match.groups()
+        family, _, _, _, _, _, _, _, weight, slope, _ = filename_match.groups()
 
         if family == condensed_name_nosp:
             family = self.condensed_family_name
```

nototools/noto_lint.py
dougfelt: I'm planning on nuking the FontProps definition and just using NotoFont.  Might rename NotoFont to FontProps, but either way I don't think I need two definitions of nearly the same struct.

PR #163, opened 2016-01-28
Tool to get HB input for each of a font's glyphs

### LINKED ISSUE #161: Include things that are not cmap'd in fontdiff
If I understand correctly, fontdiff appears to only compare what is cmap'd. 

It would be helpful to compare things that are accesible by open type feature. That is, render all the stuff you can reach by GPOS/GSUB rules as well. See discussion on #158.

This would help Google Fonts with things like broken ligatures in new versions of fonts :)

jamesgk:
Should close #161

### DIFF
--- a/nototools/hb_input.py
+++ b/nototools/hb_input.py
@@ -14,112 +14,111 @@
 
 
 
-"""Provides a functions to generate harbuzz input.
-
-The input is returned as a list of strings, suitable for passing into
-subprocess.call or something similar.
-"""
-
-
 from fontTools.ttLib import TTFont
 
 from nototools import summary
 
 
-def all_hb_inputs(path, output_dir):
-    """Generate harfbuzz inputs for all glyphs in a given font."""
+class HbInputGenerator(object):
+    """Provides functions to generate harbuzz input.
 
-    inputs = []
-    font = TTFont(path)
-    reverse_cmap = build_reverse_cmap(font)
-    glyph_names = font.getGlyphOrder()
-    for name in glyph_names:
-        inputs.append([path] + hb_input_from_name(font, name, reverse_cmap))
-    return inputs
+    The input is returned as a list of strings, suitable for passing into
+    subprocess.call or something similar.
+    """
+
+    def __init__(self, path, reverse_cmap=None):
+        self.path = path
+        self.font = TTFont(path)
+        self.reverse_cmap = reverse_cmap or build_reverse_cmap(self.font)
+
+    def all_inputs(self):
+        """Generate harfbuzz inputs for all glyphs in a given font."""
+
+        inputs = []
+        glyph_names = self.font.getGlyphOrder()
+        for name in glyph_names:
+            inputs.append([self.path] + self.input_from_name(name))
+        return inputs
+
+    def input_from_name(self, name, allow_arg=True):
+        """Given glyph name, return input to harbuzz to render this glyph, not
+        including the font path itself.
+
+        Sometimes this will simply be the glyph itself, other times a feature
+        may need to be active, or a string of glyphs used (for ligatures or
+        contextual substitutions).
+
+        If `allow_arg` is false, no feature argument will be added (used if
+        called recursively).
+        """
+
+        # see if this glyph has a simple unicode mapping
+        if name in self.reverse_cmap:
+            return [unichr(self.reverse_cmap[name])]
+
+        # nope, check the substitution features
+        gsub = self.font['GSUB'].table
+        for lookup_index, lookup in enumerate(gsub.LookupList.Lookup):
+            for st in lookup.SubTable:
+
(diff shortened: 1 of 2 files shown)


nototools/hb_input.py
```diff
@@ -0,0 +1,125 @@
...
+subprocess.call or something similar.
+"""
+
+
+from fontTools.ttLib import TTFont
+
+from nototools import summary
+
+
+def all_hb_inputs(path, output_dir):
```
dougfelt: no use of output_dir

nototools/hb_input.py
```diff
@@ -0,0 +1,125 @@
...
+    """Generate harfbuzz inputs for all glyphs in a given font."""
+
+    inputs = []
+    font = TTFont(path)
```
dougfelt: maybe pass the font in instead of creating it from the path.

nototools/hb_input.py
```diff
@@ -0,0 +1,125 @@
...
+    reverse_cmap = build_reverse_cmap(font)
+    glyph_names = font.getGlyphOrder()
+    for name in glyph_names:
+        inputs.append([path] + hb_input_from_name(font, name, reverse_cmap))
```
dougfelt: why do you prefix each sequence with the path?

nototools/hb_input.py
```diff
@@ -0,0 +1,125 @@
...
+    return inputs
+
+
+def build_reverse_cmap(font):
```
dougfelt: This assumes a 1-1 cmap.  Might warn, at least, when it's not.  As far as generating some kind of input, though, it doesn't matter I suppose.

nototools/hb_input.py
```diff
@@ -0,0 +1,125 @@
...
+    need to be active, or a string of glyphs used (for ligatures or contextual
+    substitutions).
+
+    A reverse cmap is used to check if the glyph has a unicode encoding; this
+    can be pre-built and passed in. If `allow_arg` is false, no feature argument
+    will be added (used if called recursively).
+    """
+
+    # see if this glyph has a simple unicode mapping
+    if reverse_cmap is None:
```
dougfelt: If this were a class, there'd be a natural way to save this state and not rebuild it across multiple calls.

nototools/shape_diff.py
```diff
@@ -88,27 +90,37 @@ def find_rendered_diffs(self, font_size=24, stats=None):
...
-                strin = unichr(unival)
-                subprocess.call([
-                    'hb-view', '--font-size=%d' % font_size,
-                    '--output-file=%s' % a_png, path_a, strin])
-                subprocess.call([
-                    'hb-view', '--font-size=%d' % font_size,
-                    '--output-file=%s' % b_png, path_b, strin])
+        for name in ordered_names:
+            strin, strin_b = [
+                hb_input.hb_input_from_name(f, name, self.reverse_cmap)
```
dougfelt: Doesn't this assume the same cmap in the two fonts being compared?  Might limit the usefulness of this.

nototools/hb_input.py
jamesgk: True. This is also why I was passing in the path instead of a font object. What do you think would be more useful?

nototools/hb_input.py
jamesgk: Yea, I'm not sure that's useful. I don't think it makes any difference what value is used, as long as one exists.

nototools/shape_diff.py
jamesgk: When building the reverse cmaps it will report disparity. But right now it's just taking one font's cmap where it should be taking the intersection.

jamesgk:
PTAL. Now using a class-based approach as suggested. Still not sure if the method signature of `HbInputGenerator.all_inputs` is as useful as it could be.

nototools/hb_input.py
dougfelt: If it's just a (list of?) feature tag and a string, how about a tuple of those two things.  The caller knows the path already, they passed it to you, so they can assemble the string by iterating over the tuples.

I might, for example, want to generate a text file by iterating over the strings and putting each one on a line (maybe skipping those that need a particular feature, maybe grouping by feature) and then render that with a font.  It seems cleaner to assemble this from pieces than to parse this information out of the text you're currently returning.

jamesgk:
Thanks, PTAL

behdad:
Thanks James, this is a great start!

dougfelt:
LGTM.

I'd suggest, though, that using list comprehensions to work on the pair of fonts is less clear than simply making the same method call twice, once on each element of the pair.  As it is you often extract the elements of the list/pair into separate arguments in several methods anyway.  Might as well just keep them that way all the time.

jamesgk:
Yea, that makes sense. I'll submit a separate PR later.

PR #164, opened 2016-01-30
Report error if name info looks out of date.

dougfelt:
If the name info says no preferred names are needed, and we encounter
a family that has subfamily components other than regular, bold, italic,
or bolditalic, then the name info is likely out of date.

This change also fixes a bad method name, and dumps '<none>' for fields
that are supposed to be empty (formerly it just didn't print them).


jamesgk:
LGTM

PR #166, opened 2016-02-01
Add phase 2/3 family name info files, update lint.

dougfelt:
This adds family name info files for phase 2 and 3.  Phase 2 was
generated using noto_names on the default noto directories (noto-fonts
hinted and unhinted, noto-emoji, noto-cjk) from the .ttf files; phase
3 data added files generated from the toolchain in noto-source/master_ttf.
Phase 3 files will need to be updated as we get more masters with different
variants (weight, condensed, italic, display, etc.)

Noto_lint was updated to use the --phase flag to control which name info
file to use.  Noto_names was updated to add a --phase flag to default
the filename to write the generated info to when it is not overridden.

### DIFF
--- a/nototools/family_name_info_p2.xml
+++ b/nototools/family_name_info_p2.xml
@@ -1,0 +1,123 @@
+<?xml version='1.0' encoding='utf8'?>
+<family_name_data date="2016-01-22">
+  <info family="Arimo" />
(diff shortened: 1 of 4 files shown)


jamesgk:
LGTM

PR #167, opened 2016-02-02
Change lint to use lint_cmap_reqs for cmap data, and incorporate phase.

dougfelt:
The version of _get_script_required in lint_cmap_reqs was updated to use
the phased versions of extra script characters that were added to noto_data.
An api was added to cache and get the required chars based on a NotoFont
instance and the noto phase.  Api was added to noto_data to return sets
of required/disallowed characters based on script and phase.  Noto_lint
was then updated to call lint_cmap_reqs to get the characters required
for a font.

To break a circular import, a utility function to return the 'report name'
of a script was moved from noto_lint to noto_fonts.

### DIFF
--- a/nototools/cmap_data.py
+++ b/nototools/cmap_data.py
@@ -18,7 +18,7 @@ import collections
 import datetime
 
 from nototools import lint_config
-from nototools import noto_lint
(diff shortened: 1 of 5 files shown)


jamesgk:
LGTM

PR #168, opened 2016-02-02
[fontdiff] Resize renderings to allow comparisons

### LINKED ISSUE #162: Make fontdiff compare renderings even if they aren't the same size
https://github.com/googlei18n/nototools/issues/158#issuecomment-175747707 suggests comparison is abandoned if the renders aren't the same size. For things like a ligature that has dropped part of it's contents ("fi" now renders just "i") that isn't ideal.

Resize the image to be able to provide visual comparison. Maybe just add whitespace?

jamesgk:
Should close #162

### DIFF
--- a/nototools/shape_diff.py
+++ b/nototools/shape_diff.py
@@ -23,9 +23,8 @@ magick.
 
 Neither comparison is ideal. Glyph areas can be the same even if the shapes are
 wildly different. Image comparison is usually either slow (hi-res) or inaccurate
-(lo-res), and can be easily hindered when one glyph's image is a pixel larger
-than another's. Still, these are usually useful for raising red flags and
-catching large errors.
+(lo-res). Still, these are usually useful for raising red flags and catching
+large errors.
 """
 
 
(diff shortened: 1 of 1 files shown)


dougfelt:
LGTM

nototools/shape_diff.py
```diff
@@ -116,6 +118,12 @@ def find_rendered_diffs(self, font_size=24, stats=None):
...
                 '--features=%s' % ','.join(features), path_b, text])
+
+            img_info = subprocess.check_output(['identify', a_png]).split()
+            assert img_info[0] == a_png and img_info[1] == 'PNG'
+            subprocess.call([
+                'convert', '-gravity', 'center', '-background', 'black',
+                '-extent', img_info[2], b_png, b_png])
             with open(diffs_filename, 'a') as ofile:
                 subprocess.call(
                     ['compare', '-metric', 'AE', a_png, b_png, cmp_png],
```
davelab6: I suggest using the comparison command I quoted in https://github.com/googlei18n/nototools/issues/162#issuecomment-175885431

jamesgk: I'll use that for actually saving comparison images; this is just used for getting the absolute error metric.

davelab6: Oh I see! Sorry I was just reading the code (badly! :)

PR #169, opened 2016-02-03
Various shape diff changes

jamesgk:
This makes some minor changes to the arguments and output, and allows comparison renderings to be saved.

### DIFF
--- a/nototools/fontdiff.py
+++ b/nototools/fontdiff.py
@@ -93,7 +93,7 @@ def _dump_shape_stats(stats, whitelist, out_lines, diff_type, multiple_fonts):
         print stat_format % stat
 
 
-def _gpos(path_a, path_b, out_lines, print_font=False):
+def _gpos(path_a, path_b, error_bound, out_lines, print_font=False):
     """Do a GPOS table comparison and print results.
 
     path_a and b refer to plaintext files containing ttxn output. print_font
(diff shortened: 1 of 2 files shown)


nototools/fontdiff.py
```diff
@@ -63,43 +99,41 @@ def main():
                                      'pointed to by PATH_A and PATH_B.')
     parser.add_argument('path_a', metavar='PATH_A')
     parser.add_argument('path_b', metavar='PATH_B')
-    parser.add_argument('-t', '--diff_type', default='shape',
-                        help='type of comparison to run, "shape" or "gpos" '
-                        '(defaults to "shape"), if "gpos" is provided the '
-                        'input paths should point to ttxn output')
+    parser.add_argument('-t', '--diff-type', default='area',
```
dougfelt: I think you can use 'choices' here to specify the available choices more directly.

nototools/fontdiff.py
```diff
@@ -30,20 +30,56 @@
...
+        if not multiple_fonts:
+            stat = stat[1:]
+        print stat_format % stat
+
+
+def _gpos(path_a, path_b, out_lines, print_font=False):
+    if print_font:
+        print '-- %s --' % os.path.basename(path_a)
+        print
     diff_finder = gpos_diff.GposDiffFinder(path_a, path_b, 3, out_lines)
```
dougfelt: What does the number 3 mean?

nototools/shape_diff.py
```diff
@@ -124,6 +126,16 @@ def find_rendered_diffs(self, font_size=256, stats=None):
             subprocess.call([
                 'convert', '-gravity', 'center', '-background', 'black',
                 '-extent', img_info[2], b_png, b_png])
+
+            if render_path:
+                output_png = os.path.join(render_path, name + '.png')
+                subprocess.call([
```
dougfelt: It might be useful here to document what you're doing by specifying each of these options.

dougfelt:
LGTM

PR #172, opened 2016-02-04
Update UnicodeData.txt to latest Unicode 9 draft data

roozbehp:
This should include all characters planned to be Unicode 9. No
further additions to the repretoire is expected for Unicode 9.


dougfelt:
LGTM

PR #173, opened 2016-02-09
Add tool to generate sample strings from a description file.

dougfelt:
This takes some pains to work on narrow python2 builds on mac.  It
has not been fully tested in that environment yet.

The actual task is simple in code, but the idea is to provide a
easy-to-understand text file format that linguists and other
non-programmers can edit.

### DIFF
--- a/nototools/generate_samples.py
+++ b/nototools/generate_samples.py
@@ -1,0 +1,534 @@
+#!/usr/bin/python
+#
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


nototools/generate_samples.py
```diff
@@ -0,0 +1,534 @@
...
+
+_ESCAPES = {
+    ' ': '\\x20',
+    '#': '\\x23',
+    ',': '\\x2c',
+    '-': '\\x2d',
+    '\\': '\\x5c',
+    }
+
+def _canonicalize_escapes(definition_lines):
```
jamesgk: Can we use string.translate() for this?

jamesgk:
LGTM

nototools/generate_samples.py
dougfelt: The docs say one of the arguments to string.translate is "... a 256-character string giving the translation for each character value, indexed by its ordinal."  This seems more cumbersome than what I'm doing here. In addition, I'm replacing sequences with other sequences of a different length, which translate doesn't do.  So I don't think it would work.

jamesgk: Ah, that's right. Then, what about string.replace()?

dougfelt: Yes, I think I can get that to work. I have to make sure to handle '\\ ' properly, so I have to replace '\\' before anything else, so I'll use a list of tuples instead of a map.

dougfelt:
PTAL

jamesgk:
LGTM

PR #176, opened 2016-02-17
move range parse/format fns to tool_utils.

dougfelt:
Some tools were encountering circular dependencies trying to get at these.


jamesgk:
Should we now remove these from lint_config?

dougfelt:
Yes, I was going to do that separately.  I have a bunch of stuff in flight, it's kind of a mess.

jamesgk:
OK then, this LGTM

PR #177, opened 2016-02-18
Add tool to look at block coverage of cmaps.

dougfelt:
Some tools were encountering circular dependencies trying to get at these.


jamesgk:
LGTM

PR #178, opened 2016-02-19
Add unit test for *num features

jamesgk:
This would have caught https://github.com/google/roboto/issues/71


dougfelt:
LGTM, just a few questions.

PR #179, opened 2016-02-24
Update to Unicode 9 beta files as of February 24, 2016.

roozbehp:
Various updates, but most importantly, removal of two map symbols
from Unicode 9 repretoire and more additions to script extensions.


dougfelt:
LGTM

PR #180, opened 2016-02-26
Add support for Indic properties

roozbehp:
The new functions indic_positional_category() and
indic_syllabic_category() have been added to unicode_data, that
return Indic shaping-related classes for characters.


dougfelt:
LGTM

PR #185, opened 2016-03-03
Add block info and script_to_chars fns to unicode_data.

dougfelt:
This adds utility functions for unicode blocks: one to return a
start, end range, one to return the set of chars in the range, and
one to return the list of block names in block order.

This also adds a utility function to create a dict from script
(based on script and script extensions) to chars.  The existing function
in this file returns information on only one script, and filters by
unicode version.  To do this it iterates over all characters filtering
out that don't apply.  Unfortunately, when using this function to examine
multiple scripts, you iterate over all the characters anew for each script.
It's much faster to collect all the data at once.  Also, filtering by
character age is easy to apply after the fact, so the new function does
not bother to do this.

### DIFF
--- a/nototools/unicode_data.py
+++ b/nototools/unicode_data.py
@@ -44,6 +44,8 @@ _bidi_mirroring_characters = set()
 _script_data = {}
 _script_extensions_data = {}
 _block_data = {}
+_block_range = {}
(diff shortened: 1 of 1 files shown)


jamesgk:
LGTM

PR #187, opened 2016-03-04
Fix noto_lint to handle bad font version strings.

dougfelt:
Before it even runs, lint prints out the version of the font file it is
parsing, and it was failing if the name table version didn't match the
pattern it expected.  Since the name is one of the things it checks, this
is an especially foolish thing to do...


jamesgk:
LGTM

PR #189, opened 2016-03-07
Create utility fn to bump version info in font.

dougfelt:
This was formerly rolled into the body of swat_license, which is
pretty much obsolete.  Occasionally we need to swat fonts and bump
the version data, though, so it's useful to have a utility.  At
some point we should find a better place for it.

### DIFF
--- a/nototools/swat_license.py
+++ b/nototools/swat_license.py
@@ -134,7 +134,7 @@ def get_bumped_version(ttfont, is_hinted):
   m = re.match(r'Version (\d{1,5})\.(\d{1,5})(.*)', version)
   if not m:
     print '! Could not match version string (%s)' % version
-    return False, None, None
+    return None, None
 
   major_version = m.group(1)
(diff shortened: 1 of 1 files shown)


nototools/swat_license.py
```diff
@@ -126,35 +126,15 @@ def _noto_relative_path(filepath):
...
-  print '-----\nUpdating %s' % rel_filepath
-
-  dst_file = path.join(dst_root, rel_filepath)
-
   version = names[_VERSION_ID]
   m = re.match(r'Version (\d{1,5})\.(\d{1,5})(.*)', version)
   if not m:
     print '! Could not match version string (%s)' % version
-    return
+    return False, None, None
```
jamesgk: Should this only return a 2-tuple?

dougfelt: Thanks for catching this.

jamesgk:
LGTM

PR #190, opened 2016-03-07
Add ability to output csv to mti_cmap_data tool.

dougfelt:
Because Google Spreadsheets still (!) does not provide an option to prevent
autoconversion of text to numbers, this puts 'U+' in front of each cell
so that Google Spreadsheets will leave it alone.  Then afterwards you can
mark all columns as text only in Google Spreadsheets, and replace all to remove
the leading U+.  Otherwise some of the hex is interpreted as decimal numbers,
losing the leading zeros, and some as text.  Urg.

### DIFF
--- a/nototools/mti_cmap_data.py
+++ b/nototools/mti_cmap_data.py
@@ -67,24 +67,31 @@ def get_script_to_cmap(csvdata):
   return { script: cmap for script, cmap in zip(header, data) }
 
 
-def cmap_data_from_csv(csvdata, infile=None):
+def cmap_data_from_csv(
+    csvdata, scripts=None, exclude_scripts=None, infile=None):
   args = [('infile', infile)] if infile else None
   metadata = cmap_data.create_metadata('mti_cmap_data', args)
-  tabledata = cmap_data.create_table_from_map(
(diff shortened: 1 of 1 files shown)


nototools/mti_cmap_data.py
```diff
@@ -81,24 +81,92 @@ def cmap_data_from_csv_file(csvfile):
   return cmap_data_from_csv(csvdata, csvfile)
 
 
+def csv_to_xml(csv_file, xml_file, scripts):
+  cmapdata = cmap_data_from_csv_file(csv_file)
+  if scripts:
+    print >> sys.stderr, 'not filtering by scripts when converting csv to xml'
```
jamesgk: It was a little confusing to me what this message was for at first. I might make it more explicit, like "not using 'scripts' argument when..."

jamesgk:
LGTM

nototools/mti_cmap_data.py
dougfelt: Yeah, I'll either implement this or remove the argument.

dougfelt:
There's significant additional changes, can you look at my last commit?

jamesgk:
Looks good

PR #191, opened 2016-03-08
Noto cmap reqs

dougfelt:


dougfelt:
@jamesgk sorry for the ginormous cl.  Please just look at python code/style issues.

@jungshik, can you look at the semantics to make sure the block and character assignments are reasonable?

jamesgk:
I can't even make comments on the commit! It's too big!

Line 50: `opentype_data` module is never used
Line 69: mutable default arg value
Line 149: `_op` method is never used. also, it references attr `_phase` which is never set
Line 304: shouldn't have whitespace around `=`
Line 548: `from_list` is never used
Line 574: `char_to_scripts` is never used here
Line 1145: Comment line >80 char

dougfelt:
Thanks!  I'm surprised one can't comment or view the file, it's not _that_
big.  You can view the deltas though.  Sorry again for the size.

jamesgk:
LGTM

PR #194, opened 2016-03-10
Use cmap data files in noto_lint instead of lint_cmap_reqs apis.

dougfelt:
This makes lint depend only on the generated phase2/3 cmap xml files
when validating character coverage.

### DIFF
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -371,15 +371,15 @@ def _build_cmap_dict(filename):
 _phase_2_map = None
 _phase_3_map = None
 def _get_cmap_data_for_phase(phase):
-  global _phase_2_map, _phase_3_map
-  if phase < 3:
-    if not _phase_2_map:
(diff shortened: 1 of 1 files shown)


nototools/noto_lint.py
```diff
@@ -360,6 +361,27 @@ def printable_font_versions(font):
...
+    data = cmap_data.read_cmap_data_file(filename)
+    script_to_rowdata = cmap_data.create_map_from_table(data.table)
+    return {script: frozenset(tool_utils.parse_int_ranges(rd.ranges))
+            for script, rd in script_to_rowdata.iteritems()}
+
+
+_phase_2_map = None
+_phase_3_map = None
+def _get_cmap_data_for_phase(phase):
+  global _phase_2_map, _phase_3_map
```
jamesgk: The indentation throughout most of this file seems to be 4 spaces. Though it differs throughout nototools...

jamesgk:
LGTM

PR #196, opened 2016-03-11
Noto cmap reqs

dougfelt:
Updates phase 3 cmap xml data by updating noto_cmap_reqs.  Also updates some utilities that are used to display differences between cmaps.


behdad:
What do you use default-ignorables?  Note that while HarfBuzz hides most of them, there are four that are excepted.  From hb-unicode-private.hh:

```
   * Note: While U+115F, U+1160, U+3164 and U+FFA0 are Default_Ignorable, 
   * we do NOT want to hide them, as the way Uniscribe has implemented them 
   * is with regular spacing glyphs, and that's the way fonts are made to work. 
   * As such, we make exceptions for those four. 
```

jamesgk:
LGTM

PR #197, opened 2016-03-11
Update family name info code and data.

dougfelt:
This makes a few changes:
- it adds 'omit_regular' to the data associated with a family.  formerly
  this was computed algorithmically based on whether the font was a cjk font.
  now it is in data, and when generating new data the phase affects the value
  written to the data file.
- it uses the family_id from noto_fonts as the key to the data, rather than
  using the computed preferred family name parts.  This unifies the data
  for CJK and CJK Mono fonts, which was formerly the same anyway.  The intent
  is to try to converge on a single key to use for family-related data in
  the tooling.
- it determines whether a font is cjk by looking at its attributes rather
  than whether CJK is in the key.
- it corrects for the fact that 'MONO' is now a script code and without
  special handling would get emitted in the name.
- it renames some parameter arguments from 'noto_fonts' to 'notofonts'
  to avoid colliding with the python package name.

### DIFF
--- a/nototools/family_name_info_p3.xml
+++ b/nototools/family_name_info_p3.xml
@@ -1,123 +1,120 @@
 <?xml version='1.0' encoding='utf8'?>
-<family_name_data date="2016-02-01">
-  <info family="Arimo" />
-  <info family="Cousine" />
-  <info family="Noto" use_preferred="t" use_wws="t" />
-  <info family="Noto Color Emoji" />
-  <info family="Noto Emoji" />
-  <info family="Noto Kufi Arabic" />
-  <info family="Noto Naskh Arabic" use_preferred="t" use_wws="t" />
-  <info family="Noto Nastaliq Urdu" />
-  <info family="Noto Sans" limit_original="t" use_wws="t" />
-  <info family="Noto Sans Armenian" limit_original="t" />
-  <info family="Noto Sans Avestan" />
-  <info family="Noto Sans Balinese" />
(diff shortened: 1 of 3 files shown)


nototools/family_name_info_p3.xml
```diff
@@ -1,123 +1,120 @@
...
+  <info family="sans-lgc" limit_original="t" use_wws="t" />
+  <info family="sans-limb" />
+  <info family="sans-linb" />
+  <info family="sans-lisu" />
+  <info family="sans-lyci" />
+  <info family="sans-lydi" />
+  <info family="sans-mand" />
+  <info family="sans-mlym" use_preferred="t" use_wws="t" />
+  <info family="sans-mong" />
+  <info family="sans-mono-mono" limit_original="t" use_wws="t" />
```
jamesgk: I noticed that "sans-mono-mono" is only in this file and not the other, is that intentional?

dougfelt: That's a good point.  The issue is that the phase 2 and phase 3 fonts are named differently, one is 'Noto Mono' and the other is 'Noto Sans Mono', so one has 'sans' in the family name and the other does not.  The phase2 and phase3 data aren't exactly comparable because of this, and that's a problem for comparison.  However, if one uses phase 2 data to check phase 2 fonts, and phase 3 data to check phase 3 fonts, it works as intended.  So I think it's ok to live with this for now.

jamesgk:
LGTM

PR #198, opened 2016-03-13
Command line tool with params to subset fonts using cmap.

dougfelt:
subset.py does something similar, but provides no command-line options
in its main-- it's intended to be used from a makefile, or called from
other code.

This doesn't use subset.py since it also provides an option to bump the
version number, and that's more straightforward to do before writing the
file than by reading it back in and writing it out a second time.  subset.py
is itself just a wrapper around the fonttools subset functionality so
there isn't much duplication.

### DIFF
--- a/nototools/subset_font_cmap.py
+++ b/nototools/subset_font_cmap.py
@@ -1,0 +1,112 @@
+#!/usr/bin/python
+#
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


nototools/subset_font_cmap.py
```diff
@@ -0,0 +1,112 @@
...
+
+_DEFAULT_OPTIONS = _get_default_options()
+
+_VERSION_ID = 5  # name table version string ID
+
+
+def subset_font_cmap(
+    srcname, dstname, exclude=None, include=None, bump_version=True):
+
+    opt = _DEFAULT_OPTIONS
```
jamesgk: Indentation

jamesgk:
LGTM

PR #200, opened 2016-03-15
Add cp437 data, and require the monospace font to cover it.

dougfelt:
This is a legacy PC codepage.  The expectation is that terminal emulators
and similar programs that want monospace text might want to support this
character set, so we ensure the phase 3 monospace font covers these code
points.

This also includes an update to noto_cmap_phase3.xml that reflects the fix of
an error in the PUA characters covered by the emoji font.


jamesgk:
LGTM

PR #201, opened 2016-03-15
Disable variant check except for LGC

dougfelt:
This changes the lint config syntax to allow negating string
conditions, which formerly was not supported.  This provides a simpler
alternative to disabling globally and then selectively enabling.


jamesgk:
LGTM

PR #202, opened 2016-03-16
Keep some more characters in LGC.

dougfelt:
This adds more letterlike symbols that could take a sans/serif distinction,
and the other vulgar fractions. Keeps roman numerals in Symbols based on
Jungshik's preference.

Also adds some characters to MUSIC that had been overlooked.


jamesgk:
LGTM

PR #204, opened 2016-03-17
Add some more tooling for patching Noto fonts for android use.

dougfelt:
These routines were based on code in the android repo or in the commit
history in that repo.

### DIFF
--- a/nototools/android_patches.py
+++ b/nototools/android_patches.py
@@ -32,49 +32,55 @@ from fontTools.ttLib.tables import otTables
 SRC_DIR = tool_utils.resolve_path('[tools]/packages/android')
 DST_DIR = tool_utils.resolve_path('[tools]/packages/android-patched')
 
-def _patch_hyphen_armenian_ethiopic():
-  """Add hyphen-minus glyphs to Armenian and Ethiopic fonts.
+def _patch_hyphen():
+  """Add hyphen-minus glyphs to fonts that need it.
 
-  This is to enable Armenian and Amharic langauges to be hyphenated
-  properly, since Minikin's itemizer currently shows tofus if a an
-  automatically hyphenated word is displated in a font that has neither
-  HYPHEN nor HYPHEN-MINUS.
+  This is to enable languages to be hyphenated properly,
+  since Minikin's itemizer currently shows tofus if an
+  automatically hyphenated word is displated in a font
+  that has neither HYPHEN nor HYPHEN-MINUS.
 
-  In practice only U+002D HYPHEN-MINUS is added, since Noto LGC fonts
+  The list of font names comes from LANG_TO_SCRIPT in
+  tools/font/fontchain_lint.py.
+
+  (In practice only U+002D HYPHEN-MINUS is added, since Noto LGC fonts
   don't have U+2010 HYPHEN.)
 
   Bug: 21570828"""
 
-  FONTS = [
-      'NotoSansArmenian-Regular.ttf',
-      'NotoSansArmenian-Bold.ttf',
-      'NotoSerifArmenian-Regular.ttf',
-      'NotoSerifArmenian-Bold.ttf',
-      'NotoSansEthiopic-Regular.ttf',
-      'NotoSansEthiopic-Bold.ttf',
-  ]
+  # Names of fonts for which Android requires a hyphen.
+  # This list omits Japanese and Korean.
+  script_names = [
+      'Armenian', 'Ethiopic', 'Bengali', 'Gujarati', 'Devanagari',
+      'Kannada', 'Malayalam', 'Oriya', 'Gurmukhi', 'Tamil', 'Telugu']
 
   HYPHENS = {0x002D, 0x2010}
 
-  for font_name in FONTS:
-    lgc_font_name = (
-        font_name.replace('Armenian', '')
-        .replace('Ethiopic', ''))
+  for sn in script_names:
+    globexp = path.join(SRC_DIR, 'Noto*%s-*.ttf' % sn)
+    fonts = glob.glob(globexp)
+    if not fonts:
+      raise ValueError('could not match ' + globexp)
+    fonts = [path.basename(f) for f in fonts]
+    for font_name in fonts:
+      lgc_font_name = font_name.replace(sn, '')
 
-    font_file = path.join(SRC_DIR, font_name)
-    lgc_font_file = path.join(SRC_DIR, lgc_font_name)
+      font_file = path.join(SRC_DIR, font_name)
+      lgc_font_file = path.join(SRC_DIR, lgc_font_name)
 
-    chars_to_add = (
-        (HYPHENS - coverage.character_set(font_file))
-        & coverage.character_set(lgc_font_file))
+      chars_to_add = (
+          (HYPHENS - coverage.character_set(font_file))
+          & coverage.character_set(lgc_font_file))
 
-    if chars_to_add:
-      print 'patch hyphens', font_name
-      merger.merge_chars_from_bank(
-          path.join(SRC_DIR, font_name),
-          path.join(SRC_DIR, lgc_font_name),
-          path.join(DST_DIR, font_name),
-          chars_to_add)
+      if chars_to_add:
+        print 'patch hyphens', font_name
+        merger.merge_chars_from_bank(
+            path.join(SRC_DIR, font_name),
+            path.join(SRC_DIR, lgc_font_name),
+            path.join(DST_DIR, font_name),
+            chars_to_add)
+      else:
+        print 'already have hyphens', font_name
 
 
 def _remove_cjk_emoji():
@@ -266,7 +272,7 @@ def _subset_symbols():
 
 def main():
   tool_utils.ensure_dir_exists(DST_DIR)
-  _patch_hyphen_armenian_ethiopic()
+  _patch_hyphen()
   # TODO: first unpack from ttc, then rebuild
   # _remove_cjk_emoji()
   _subset_symbols()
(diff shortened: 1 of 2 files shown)


nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+# limitations under the License.
+
+"""Patches for Android versions of Noto fonts."""
+
+import glob
+from os import path
+
+from nototools import subset
+from nototools import coverage
+from nototools import fix_khmer_and_lao_coverage as merger
```
roozbehp: Probably a good idea to move merge_chars_from_bank from fix_khmer... to some other module that both scripts can share.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+from nototools import unicode_data
+
+from fontTools import ttLib
+from fontTools.ttLib.tables import otTables
+
+SRC_DIR = tool_utils.resolve_path('[tools]/packages/android')
+DST_DIR = tool_utils.resolve_path('[tools]/packages/android-patched')
+
+def _patch_serif_lgc():
+  """Delete smiley and playing card suits from Noto Serif LGC.
```
roozbehp: No longer necessary, since Unicode specifies them as default text and Android now supports variation selectors.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+      merger.merge_chars_from_bank(
+          path.join(SRC_DIR, font_name),
+          path.join(SRC_DIR, lgc_font_name),
+          path.join(DST_DIR, font_name),
+          chars_to_add)
+
+
+# Characters we have decided we are doing as emoji-style in Android,
+# despite UTR#51's recommendation
+ANDROID_EMOJI = [
```
roozbehp: This is also going away. We are making these text style.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+  doesn't cause a difference, as they are not presently supported in
+  Minikin.
+
+  The Korean font doesn't change in coverage, but seems to get some
+  size reduction being passed through fontTools, so it's included in
+  the patch.
+
+  (1c4749e20391a4)
+  ----
+  ### NOTE: Roozbeh says (2016/03/10) this is no longer needed.
```
roozbehp: Yes. The twleve should now default to text.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+  """(two fixes)
+  ---
+  Remove default emoji characters from CJK fonts.
+
+  Twenty-six characters that Unicode Technical Report #51 "Unicode
+  Emoji" defines as defaulting to emoji styles used to be displayed as
+  black and white ("text" style) before this. This patch removes those
+  characters from Noto CJK fonts, so they get displayed as color.
+
+  As a side effect of the subsetting process, some ideographic
```
roozbehp: Please make sure the variation sequences are not lost anymore, now that we support them.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+    0x2132, # Ⅎ TURNED CAPITAL F
+    0x213B, # ℻ FACSIMILE SIGN
+    0x214D, # ⅍ AKTIESELSKAB
+    0x214F, # ⅏ SYMBOL FOR SAMARITAN SOURCE
+}
+
+# default emoji characters in the BMP, based on
+# http://www.unicode.org/draft/Public/emoji/1.0/emoji-data.txt
+# We exclude these, so we don't block color emoji.
+BMP_DEFAULT_EMOJI = {
```
roozbehp: Please get this from emoji-data.txt instead of hard-coding it.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+    0x2B24, # ⬤ BLACK LARGE CIRCLE
+    0x2B2E, # ⬮ BLACK VERTICAL ELLIPSE
+    0x2B2F, # ⬯ WHITE VERTICAL ELLIPSE
+    0x2B56, # ⭖ HEAVY OVAL WITH OVAL INSIDE
+    0x2B57, # ⭗ HEAVY CIRCLE WITH CIRCLE INSIDE
+    0x2B58, # ⭘ HEAVY CIRCLE
+    0x2B59, # ⭙ HEAVY CIRCLED SALTIRE
+}
+
+# letter-based characters, provided by Roboto
```
roozbehp: Add a TODO here to see if we need to change this subset based on Noto Serif coverage (so the serif fallback chain would support them).

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+    0x2B55, # ⭕ HEAVY LARGE CIRCLE
+}
+
+
+def _subset_symbols():
+  """Subset Noto Sans Symbols in a curated way.
+
+  Noto Sans Symbols is now subsetted in a curated way. Changes include:
+
+  * Currency symbols now included in Roboto are removed.
```
roozbehp: Same as above (regarding Noto Serif)

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+  for first, last, _ in unicode_data._parse_code_ranges(BLOCKS_TO_INCLUDE):
+    target_coverage.update(range(first, last+1))
+
+  # Add one-off characters
+  target_coverage |= ONE_OFF_ADDITIONS
+  # Remove characters preferably coming from Roboto
+  target_coverage -= LETTERLIKE_CHARS_IN_ROBOTO
+  # Remove characters that are supposed to default to emoji
+
+  # According to Roozbeh, CJK is not available on all platforms, so even
```
roozbehp: Please update after removing ANDROID_EMOJI altogether.

nototools/android_patches.py
roozbehp: But the ones starting from line 159 should still be removed.

nototools/android_patches.py
```diff
@@ -0,0 +1,431 @@
...
+  According to the OpenType layout specifications, Khmer and Lao fonts
+  need to support DOTTED CIRCLE and ZWSP, while Khmer addtionally needs
+  to support ZWNJ and ZWJ.  Otherwise, vowels without base letters
+  would be displyed without a dotted circle, among other potential
+  issues (https://code.google.com/p/noto/issues/detail?id=4).
+
+  This adds glyphs from Noto Sans LGC for ZWJ, ZWNJ, and DOTTED CIRCLE
+  to the Khmer fonts, and ZWSP and DOTTED CIRCLE to the Lao fonts.
+  """
+  # There's no difference between our Khmer/Lao and Android's, so
```
roozbehp: Then remove and add to lint please.

nototools/android_patches.py
dougfelt: Will think about this.  That script can probably just be deleted and the merge fn moved to this one.

nototools/android_patches.py
dougfelt: Deleted.

nototools/android_patches.py
dougfelt: Not used, meant to delete, will do so.

nototools/android_patches.py
dougfelt: No variation sequences in the CJK font I looked at.

nototools/android_patches.py
dougfelt: OK.  Not quite sure what BMP is meant to imply, it doesn't seem to require a limit to BMP chars.

nototools/android_patches.py
dougfelt: Removed.  Lint already requires these characters.

nototools/android_patches.py
```diff
@@ -0,0 +1,276 @@
...
+from nototools import tool_utils
+from nototools import unicode_data
+
+from fontTools import ttLib
+from fontTools.ttLib.tables import otTables
+
+SRC_DIR = tool_utils.resolve_path('[tools]/packages/android')
+DST_DIR = tool_utils.resolve_path('[tools]/packages/android-patched')
+
+def _patch_hyphen_armenian_ethiopic():
```
jungshik: This need to be expanded to other script fonts that need to have a hyphen.

dougfelt: Yeah, I know.  Where is the code that defines what scripts these are?  Is
it in minikin somewhere?

dougfelt:
Updated script to check more fonts for hyphens.  Only Armenian and Ethiopic get hyphens added, as it turns out.

nototools/android_patches.py
```diff
@@ -32,49 +32,55 @@
...
-      'NotoSerifArmenian-Regular.ttf',
-      'NotoSerifArmenian-Bold.ttf',
-      'NotoSansEthiopic-Regular.ttf',
-      'NotoSansEthiopic-Bold.ttf',
-  ]
+  # Names of fonts for which Android requires a hyphen.
+  # This list omits Japanese and Korean.
+  script_names = [
+      'Armenian', 'Ethiopic', 'Bengali', 'Gujarati', 'Devanagari',
+      'Kannada', 'Malayalam', 'Oriya', 'Gurmukhi', 'Tamil', 'Telugu']
```
dougfelt: Were you going to update the LGC fonts with the prime glyphs from the fonts
Marek made, or should I?

jungshik: Just for the record, prime glyphs (googlei18n/noto-fonts#510) were taken care of in Noto Serif.

dougfelt:
@jungshik, can you look at this and let me know if it is ok to pull?

PR #206, opened 2016-03-17
Use unicodedata2 if available.

### LINKED ISSUE #205: unicode_data.py to use unicodedata2
I see that [unicode_data.py](https://github.com/googlei18n/nototools/blob/master/nototools/unicode_data.py) is doing: 
```
import unicodedata  # Python's internal library
```

It would be more efficient to do: 
```
try: 
  import unicodedata2 as unicodedata
except ImportError: 
  import unicodedata
```

`unicodedata2` is a C++ Python 2.7-compatible "unicodedata" replacement (Apache2-licensed) which is currently at Unicode 8.0. So the "inefficient" lookups done in pure Python could be sped up if that library was used:
https://github.com/mikekap/unicodedata2
https://pypi.python.org/pypi/unicodedata2

dougfelt:
Fixes #205

### DIFF
--- a/nototools/unicode_data.py
+++ b/nototools/unicode_data.py
@@ -31,7 +31,11 @@ import collections
 import os
 from os import path
 import re
-import unicodedata  # Python's internal library
+import sys
+try:
+  import unicodedata2 as unicodedata  # Unicode 8 compliant native lib
+except ImportError:
+  import unicodedata  # Python's internal library
 
 
 _data_is_loaded = False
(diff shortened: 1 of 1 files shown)


nototools/unicode_data.py
```diff
@@ -31,7 +31,11 @@
 import os
 from os import path
 import re
-import unicodedata  # Python's internal library
+import sys
+try:
+  import unicodedata2 as unicodedata  # Unicode 8 compliant c++ lib
```
anthrotype: I'm not sure where @twardoch got the impression that it's written in C++...
To me it seems like plain C:
https://github.com/mikekap/unicodedata2/blob/master/unicodedata2/unicodedata.c

jamesgk:
LGTM

PR #207, opened 2016-03-18
Add additional checks to noto cmap reqs.

dougfelt:
Roozbeh suggested adding a check to ensure that if a bidi mirroring char
is in a font, its mirrored sibling is also in the font.  This adds that
to the cmap generation code.

I'd also noticed that I hadn't explicitly checked the opentype_data
list of special characters.  While this could have been incorporated
into the list of script required characters, it seems more correct to
keep the opentype requirements separate from the others which are mostly
based on unicode data.

As it turns out, this does not impact the resulting cmaps (we were already
covered on both counts) so this commit does not include an update to the
cmap data.

This also add a utility fn to unicode_data to return the set of all
mirrored characters.

### DIFF
--- a/nototools/noto_cmap_reqs.py
+++ b/nototools/noto_cmap_reqs.py
@@ -47,6 +47,7 @@ from nototools import cmap_data
 from nototools import compare_cmap_data
 from nototools import collect_cldr_punct
 from nototools import noto_data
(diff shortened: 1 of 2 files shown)


jamesgk:
LGTM

PR #209, opened 2016-03-22
[fontdiff] Add option to compare GSUB

jamesgk:
This is pretty basic and isn't very useful if glyph names don't match,
but it's better than nothing.


dougfelt:
LGTM

PR #210, opened 2016-03-24
Load cmap and name data files using full path names.

dougfelt:
Previously only filenames were used, so loading depended on the user's
current directory.  This should let lint be run from anyplace.

### DIFF
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -364,7 +364,7 @@ def printable_font_versions(font):
 def _build_cmap_dict(filename):
     tooldir = tool_utils.resolve_path('[tools]/nototools')
     data = cmap_data.read_cmap_data_file(
-        path.join(tooldir, filepath))
+        path.join(tooldir, filename))
     script_to_rowdata = cmap_data.create_map_from_table(data.table)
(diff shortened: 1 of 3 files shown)


nototools/noto_lint.py
```diff
@@ -362,7 +362,9 @@ def printable_font_versions(font):
 
 
 def _build_cmap_dict(filename):
-    data = cmap_data.read_cmap_data_file(filename)
+    tooldir = tool_utils.resolve_path('[tools]/nototools')
+    data = cmap_data.read_cmap_data_file(
+        path.join(tooldir, filepath))
```
jungshik: Just a thought. Would it be worth adding a helper method (to tool_utils)  for constructing a full path (that combines resolve_path and path.join above) ?

jungshik:
Thanks. LGTM. Feel free to merge with or without addressing a minor comment above.

nototools/noto_lint.py
dougfelt: This doesn't come up that much, so it doesn't seem worth it to me.  I could have modified the file names to include the complete path too, which would also have worked.

PR #211, opened 2016-03-24
Emoji name data minor fixes

dougfelt:


jamesgk:
I'm still of the opinion that we should use the json Python module, even for a trivial task. But otherwise LGTM.

PR #213, opened 2016-03-30
Update cmap data tooling to handle 'fallback' codepoints.

dougfelt:
We are suggesting some additional codepoints for fonts based on possible
typographic interaction, and asking MTI to examine whether these additional
codepoints can be omitted from the font and supported via fallback
from the LGC fonts.  We communicate with them using a spreadsheet, when
they have examined a font they mark the font name with an asterisk, and
additionally append an asterisk after each codepoint they feel does not
need to be in the font.

Formerly the tooling did not support this additional data, it only
supported a list of cps for each font.  This makes changes to support an
optional additional list of 'fallback cps'.

cmap_data is modified to read/write 6 columns of data instead of 4, the
two additional columns are the count of 'fallback' code points and the
text representation of these ranges.  When there is no data on fallback
code points, the count is -1.  The data in the two ranges should not
overlap, this however is not verified.

mti_cmap_data is modified to read and write the new data in .csv form,
converting it to the canonical xml form.

compare_cmap_data compares two data files (either of which might or might
not have the additional columns on fallback cps) and reports differences
between them, including differences between the fallback sets, and also
reports on cps that move between standard and fallback sets.

noto_cmap_reqs is modified to merge the computed cmap data (which
does not change) with fallback cmap data from a separate file.

noto_lint does not change, since fallback cps are removed from the standard
cps, noto_lint just sees a smaller set of required cps.

### DIFF
--- a/nototools/noto_cmap_reqs.py
+++ b/nototools/noto_cmap_reqs.py
@@ -2615,10 +2615,32 @@ def build_script_to_chars(log_level):
   return cmap_ops.create_script_to_chars()
 
 
+def _merge_fallback_chars(script_to_chars, srcfile):
+  xtra_cmap_data = cmap_data.read_cmap_data_file(srcfile)
+  xtra_rowdata = cmap_data.create_map_from_table(xtra_cmap_data.table)
+  merged_cmap = {}
+  for script in sorted(script_to_chars):
+    cmap = script_to_chars[script]
+    xcmap = None
+    if script in xtra_rowdata:
+      rowdata = xtra_rowdata[script]
+      xcount = int(getattr(rowdata, 'xcount', -1))
+      if xcount != -1:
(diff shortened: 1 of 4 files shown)


nototools/noto_cmap_reqs.py
```diff
@@ -2615,12 +2615,34 @@ def build_script_to_chars(log_level):
...
+  return merged_cmap
+
+
 def _get_cmap_data(script_to_chars):
   metadata = cmap_data.create_metadata('noto_cmap_reqs', [])
   tabledata = cmap_data.create_table_from_map(script_to_chars)
   return cmap_data.CmapData(metadata, tabledata)
 
 
+
```
jamesgk: Accidental new line?

jamesgk:
LGTM

PR #214, opened 2016-03-31
Update Arab, Hebr, Thai fallback cmap data.

dougfelt:
Also update tool to generate information on merge file when one is used.

This version of the cmap data includes columns with information on cmap fallbacks, taken from parsing the mti spreadsheet .csv file.


jamesgk:
LGTM

PR #215, opened 2016-04-02
Add utility to extract and represent emoji ordering information.

dougfelt:
This parses either a .csv file or emojiOrdering.txt file that contain
information about grouping and ordering of emoji in categories and
subcategories.  It defines a class that provides access to this
data.

### DIFF
--- a/nototools/emoji_ordering.py
+++ b/nototools/emoji_ordering.py
@@ -1,0 +1,269 @@
+#!/usr/bin/python
+#
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


dougfelt:
ping

nototools/emoji_ordering.py
```diff
@@ -0,0 +1,269 @@
...
+
+import argparse
+import codecs
+import collections
+from itertools import chain
+from os import path
+import re
+import sys
+
+def _generate_emoji_to_cat_tuple(category_odict):
```
jamesgk: Could/should this be a method of EmojiOrdering?

jamesgk:
LGTM

nototools/emoji_ordering.py
dougfelt: Could.  The line between what's in the class and what's not is a little fuzzy.  The data structure, for example, is built outside the class by the parsing code.  The conversion of the data structure to the inverse map seemed to me more like a utility function that could be used elsewhere (e.g. to verify the original data structure doesn't contain multiple mappings for the same sequence, which I currently don't do) so I left it out of the class.  I'll probably leave it as is.

PR #217, opened 2016-04-08
Quiet warning/log messages

dougfelt:
Now uses a single fn for all warning/log messages, including those
that somehow had been missed earlier and so were still noisy.


jamesgk:
LGTM

PR #219, opened 2016-04-12
Update name data to support new phase3 fonts.

dougfelt:
- Also adds support for parsing font file names with 'SemiCondensed',
  'ExtraCondensed' variants.  Name expectations are not updated yet.


dougfelt:
@marekjez86 this should take care of parsing issues.  The expectations for how the name tables should read are not updated yet (they expect 'SemiCondensed' and 'ExtraCondensed' where 'Condensed' appears, but the fonts have a space between 'Extra' and 'Condensed', and of course the fonts still generate the parts of the name in different locations from where lint expects them.

jamesgk:
LGTM

PR #220, opened 2016-04-12
Add/enhance tool utils.

dougfelt:
Adds a function set up logging using a log level name, and optionally
quiet some ttx/fontTools messages that log at info level.

Adds an option to clean a directory in ensure_dir_exists.

Adds a utility to shorten a path to a cwd-relative or noto keyname
style pathname, to make logging/debugging strings a bit shorter.

### DIFF
--- a/nototools/tool_utils.py
+++ b/nototools/tool_utils.py
@@ -353,17 +353,8 @@ def setup_logging(loglevel, quiet_ttx=True):
            'error, critical, or a numeric value')
     return
   logging.basicConfig(level=loglevel)
-  logger = logging.getLogger()
 
-  if quiet_ttx:
-    class QuietFontToolsFilter(object):
-      def __init__(self, val):
-        self._ttx_val = val
-      def filter(self, log_record):
-        if (log_record.name in ['fontTools.misc.xmlReader', 'fontTools.ttLib']
-            and log_record.levelno == 20
-            and self._ttx_val >= 20):
-          return 0
-        return 1
-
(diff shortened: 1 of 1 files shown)


nototools/tool_utils.py
```diff
@@ -303,3 +332,38 @@ def emit():
...
+
+def setup_logging(loglevel, quiet_ttx=True):
+  """Set up logging to stream to stdout.
+
+  The loglevel is a logging level name or a level value (int or string).
+
+  ttx/fontTools uses 'info' to report when it is reading/writing tables,
+  but when we want 'info' in our own tools we usually don't want this detail.
+  When quiet_ttx is true, set up logging to treat 'info' logs from
+  fontTools misc.xmlReader and ttLib as though they were at level 19."""
```
anthrotype: @dougfelt you don't need a `Filter` for this. If you simply want to mute, say, `info()` messages from one logger, you can simply set that specific logger's `level` to a value that is higher than the root logger's level.
In fonttools, we follow the convention to have one logger instance per module, and name it after the module itself.

```python
if loglevel == logging.INFO:
    for module in ('fontTools.misc.xmlReader', 'fontTools.ttLib'):
        logger = logging.getLogger(module)
        logger.setLevel(loglevel + 1)
```

I agree that ttLib and xmlReader are a bit noisy (they were like that already, so I kept them as is). But at least now that we use `logging` clients can mute them more easily.

jamesgk:
LGTM

anthrotype:
👍

PR #221, opened 2016-04-20
Add additional symbols that need to be in subsetted2 and not subsetted.

dougfelt:
- also adds some code to format/generate the set of codepoints for the
  source file.


jamesgk:
LGTM

PR #223, opened 2016-04-27
Update patch tool for android to put new Japanese TV symbols into sub…

dougfelt:
…setted.


nototools/android_patches.py
```diff
@@ -214,6 +214,10 @@ def _remove_from_cmap(infile, outfile, exclude=[]):
     0x2764,  # ❤ HEAVY BLACK HEART
 }
 
+# TV symbols, see issue #557
```
jamesgk: Since this isn't a nototools bug, maybe we should link directly to it.

jamesgk:
LGTM

PR #225, opened 2016-04-27
Add os/2 PANOSE checking to lint.

dougfelt:
We don't care much about panose, but do care a little.  This only checks a
few things:
- bFamilyType is 2
- bSerifStyle is 0, or 2-10 for serif, or 11 and up for sans.  We allow
  0 'any' if the font can be used as either.
- bProportion is 9 for monospace, or 2-4 for non-monospace.  A value of 0
  (any) doesn't seem to make sense, that would mean a font could be both
  monospace and non-monospace, so we disallow it.

The general approach is that non-LGC fonts should have values that match those
of their LGC counterparts.  In particular we do not want '1 1 1 1 1 1 1 1 1 1'
for non-LGC fonts.

### DIFF
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -1051,6 +1051,38 @@ def check_font(font_props, filename_error,
 
             check_ul_unicode_range()
 
+            if os2_table.panose.bFamilyType != 2:
+              warn("head/os2/panose/family", "OS/2",
+                   "Panose family value is %s but expected 2" %
+                   os2_table.panose.bSerifStyle)
+
+            expect_serif = noto_font.style == 'Serif' or noto_font.family in [
+                'Cousine', 'Tinos']
+            expected_serif_range_str = '2-10' if expect_serif else '11-15'
+            serif_val = os2_table.panose.bSerifStyle
+            is_serif = 1 < serif_val < 11
+            if serif_val == 1:
+              warn("head/os2/panose/serif", "OS/2",
+                   "Panose serif value is no_fit (1) but expected 0, %s" %
+                   expected_serif_range_str)
+            elif serif_val != 0 and expect_serif != is_serif:
+              warn("head/os2/panose/serif", "OS/2",
+                   "Panose serif value is %s but expected %s" %
+                   (serif_val, expected_serif_range_str))
+
+            expect_mono = noto_font.is_mono or noto_font.family == 'Cousine'
+            expect_mono_range_str = '9' if expect_mono else '2-4'
+            proportion_val = os2_table.panose.bProportion
+            is_mono = proportion_val == 9
+            if proportion_val <= 1:
+              warn("head/os2/panose/proportion", "OS/2",
+                   "Panose proportion value is any (0) or no_fit (1) but "
+                   "expected %s" % expect_mono_range_str)
+            elif expect_mono != is_mono:
+              warn("head/os2/panose/proportion", "OS/2",
+                   "Panose proportion value is %s but expected %s" %
+                   (proportion_val, expect_mono_range_str))
+
 
     def check_vertical_limits():
         if 'glyf' not in font:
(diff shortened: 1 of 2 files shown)


nototools/noto_lint.py
```diff
@@ -1051,6 +1051,38 @@ def check_ul_unicode_range():
...
+                   os2_table.panose.bSerifStyle)
+
+            expect_serif = noto_font.style == 'Serif' or noto_font.family in [
+                'Cousine', 'Tinos']
+            expected_serif_range_str = '2-10' if expect_serif else '11-15'
+            serif_val = os2_table.panose.bSerifStyle
+            is_serif = 1 < serif_val < 11
+            if serif_val == 1:
+              warn("head/os2/panose/serif", "OS/2",
+                   "Panose serif value is no_fit (1) but expected 0, %s" %
```
jungshik: I wonder which is better,  "0, 2-10" or "0 or 2-10"?  (2-10 can be 11-15).

nototools/noto_lint.py
```diff
@@ -1051,6 +1051,38 @@ def check_ul_unicode_range():
...
+                   expected_serif_range_str)
+            elif serif_val != 0 and expect_serif != is_serif:
+              warn("head/os2/panose/serif", "OS/2",
+                   "Panose serif value is %s but expected %s" %
+                   (serif_val, expected_serif_range_str))
+
+            expect_mono = noto_font.is_mono or noto_font.family == 'Cousine'
+            expect_mono_range_str = '9' if expect_mono else '2-4'
```
jungshik: How about 'Condensed' (6 or 8)?  

nit:  wouldn't it be better to use |expect_proportion_range| to be paired with |proportion_val| ? Well, we still have |expect_mono|. ....

jungshik:
Do you plan to add 'Condensed' later? (see my comment to bProportion)?

nototools/noto_lint.py
dougfelt: Yes, 'or' is probably better.

nototools/noto_lint.py
dougfelt: It seems pretty likely to me that a test for bProportion == 9 is actually used somewhere to identify a monospace font.  It's not clear to me that anyone uses the other proportion values to identify a condensed/wide font.  Do you know of some code that does?

davelab6: On 27 April 2016 at 20:07, dougfelt <notifications@github.com> wrote:

> It seems pretty likely to me that a test for bProportion == 9 is actually
> used somewhere to identify a monospace font.
>
Windows likes monospace fonts to have bProportion set correctly

> It's not clear to me that anyone uses the other proportion values to
> identify a condensed/wide font. Do you know of some code that does?
>
Also, if PANOSE values don't match across the fonts in a family, Windows
can misbehave

jungshik: > It seems pretty likely to me that a test for bProportion == 9 is actually
> used somewhere to identify a monospace font.

In addition to Windows, Skia does that (I sent you a link to the code where it does in an email thread on the topic). 

> It's not clear to me that anyone uses the other proportion values to identify a condensed/wide font. 
> Do you know of some code that does?

I don't know any (other than monospace), but I thought it's just a bit tidier to mark 'condensed' with 6 or 8. Unlike other attributes, 'condensed' is a rather clear distinction, isn't it?

dougfelt: With variation fonts that can have interpolated widths this gets a bit murkier.  Right now I'm keying off the font name to determine the expectations.  I guess as long as the values correspond to the name it's ok.

I'd prefer to submit this as is, then submit the updates to lint so that it can parse the font names, then look at what kinds of issues arise with condensed/semicondensed fonts and fix this then.

dougfelt: @jungshik, the link you sent showed use of the Style value but not the Proportion value.  Maybe Skia uses it elsewhere but I didn't see it referenced in the Chromium Skia code.

jungshik: @dougfelt You're right. I misremembered that. Anyway, if your preference is to just go ahead, it's fine with me. LGTM.

PR #227, opened 2016-04-29
Add Lao Kip (currency) to Lao font.

dougfelt:
This looks like the only script-specific currency symbol we need to add.


jungshik:
LGTM

PR #230, opened 2016-05-19
Lint namedata update

dougfelt:
Updates the name data xml files for phase 2 and 3.  Phase 2 had omitted serif Devanagari, this fixes a bug reported by MTI via email.  To prevent further hiccups like this in phase 3, I decided to add all the planned phase 3 fonts to the data.

To do this I created a data file based on the planning spreadsheet page that lists the planned masters, wrote a tool that processes this and generates a list of filenames, and updated the name generation tool to take a list of names as input in addition to individual names and dirs on the command line, and ran the tool to generate the new phase 3 name data file.  I also updated noto_fonts to allow it to parse two new font file names it wasn't prepared for.

### DIFF
--- a/nototools/check_familyname_and_styles.py
+++ b/nototools/check_familyname_and_styles.py
@@ -1,0 +1,174 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Process a family description file.
+
+You can check the file, and generate a list of font file names from it.
+This list can be passed to noto_names to generate the name data.
+
+The file is a list of Noto family names (see noto_fonts.py) interspersed with
+definitions of what WWS style combinations apply to that file.  See
+_get_stylenames() for the format.  Each style definition applies to each
+following family until the next style definition."""
+
+import argparse
+import re
+
+from nototools import noto_fonts
+
+_style_re = re.compile(r'--\s+(.*)\s+--')
+_extended_style_re = re.compile(r'^([TRBH]+)(?:/([CR]+)(?:/([RI]+))?)?$')
+
+# Below we use the longest names we intend, so that the noto_names code can
+# identify which families need extra short abbreviations.  The style of
+# abbreviation is based on the longest names in the family.
+
+_WEIGHT_NAMES = {
+    'T': 'Thin',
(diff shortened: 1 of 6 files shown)


nototools/check_familyname_and_styles.py
```diff
@@ -0,0 +1,174 @@
...
+      if m:
+        styles = m.group(1)
+        continue
+      assert name[0] != '-'
+      fn(name, styles)
+
+
+def check_familynames(namefile):
+  passed = [True]
+  def fn(namefile, styles):
```
jamesgk: Would `namefile` here more accurately be named `name`?

nototools/check_familyname_and_styles.py
```diff
@@ -0,0 +1,174 @@
...
+  encoding of the styles, see _get_stylenames."""
+  stylenames = _get_stylenames(styles)
+  return [name + '-' + s + '.ttf' for s in stylenames]
+
+
+def _for_all_familynames(namefile, fn):
+  """Call fn passing the family name and style descriptor for
+  all families in namefile. '#' is a comment to eol, blank lines are
+  ignored."""
+  styles = 'RB'
```
jamesgk: Is it worth including an initial value for `styles`? It should definitely be set before any names are processed, right?

nototools/check_familyname_and_styles.py
```diff
@@ -0,0 +1,174 @@
...
+        name = name[:ix]
+      name = name.strip()
+      if not name:
+        continue
+
+      m = _style_re.match(name)
+      if m:
+        styles = m.group(1)
+        continue
+      assert name[0] != '-'
```
jamesgk: Might want to note what this is checking for.

nototools/check_familyname_and_styles.py
```diff
@@ -0,0 +1,174 @@
...
+  (master-ish) style names encoded as weights / widths/ italic, where weights
+  is any of 'T', 'R', 'B', or 'H', widths any of 'C' or 'R', and italic 'I'.
+  If there's not an italic then the italic is omitted, if there's only
+  regular width and no italic then widths are omitted."""
+  m = _extended_style_re.match(styles)
+  assert m
+  weights = m.group(1)
+  widths = m.group(2) or 'R'
+  slopes = m.group(3) or 'R'
+  # print '%s: %s, %s, %s' % (styles, weights, widths, slopes)
```
jamesgk: Did you want to leave this in?

jamesgk:
LGTM with comments

nototools/check_familyname_and_styles.py
dougfelt: D'oh, sure would be.

nototools/check_familyname_and_styles.py
dougfelt: Well, it  was either setting the initial value, or failing with an error if it is not set before the first filename.  I guess it should be explicitly set, you're right.

nototools/check_familyname_and_styles.py
dougfelt: Yes.

nototools/check_familyname_and_styles.py
dougfelt: Removed.

PR #232, opened 2016-05-25
Rename fontdiff to otfdiff

jamesgk:
This is in anticipation of a new standalone fontdiff tool, which we should merge this into.

cc @davelab6, I hope this won't break anything. The thinking is that fontdiff is basically the better name but we should move forward with https://github.com/googlei18n/fontdiff and merge any missing functionality from here to there (off the top of my head I could think of diff sorting / filtering and some script to generate input based on a font's coverage).


davelab6:
OTF generally refers to OpenType-CFF flavor, but the nototools diff'er will
work on TTFs, right? In this case I suggest "notodiff" or "notofontdiff" :)

behdad:
notodiff sounds best for now.

PR #234, opened 2016-05-31
Make it possible to compare two scripts in the same file.

dougfelt:
The previous script comparison code compared the same script across
two files.  This change makes it easier to compare, say, Arab and Aran,
or LGC and MONO.

- also adds support for 'MONO' as a pseudo-script code.

### DIFF
--- a/nototools/compare_cmap_data.py
+++ b/nototools/compare_cmap_data.py
@@ -33,9 +33,18 @@ def title_from_metadata(metadata):
   return ' '.join(items)
 
 
-def compare_cmaps(
-    base_cps, target_cps, only_cps, except_cps, no_additions, no_removals):
+CompareOptions = collections.namedtuple(
+    'CompareOptions', 'only_cps, except_cps, no_additions, no_removals')
+
+
+_DEFAULT_COMPARE_OPTIONS = CompareOptions(None, None, False, False)
+
+
+def compare_cmaps(base_cps, target_cps, opts=None):
   """Returns a tuple of added, removed."""
+  if opts == None:
+    opts = _DEFAULT_COMPARE_OPTIONS
+  only_cps, except_cps, no_additions, no_removals = opts
   if only_cps:
     base_cps &= only_cps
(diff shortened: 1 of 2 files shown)


nototools/compare_cmap_data.py
```diff
@@ -62,33 +84,54 @@ def compare_cmap_data(base_cmap_data, target_cmap_data, scripts, cps,
...
+
+
+def compare_interscript_data(base_cmap_data, scripts, opts=None):
+  result = {}
+  base_map = cmap_data.create_map_from_table(base_cmap_data.table)
+  for i in range(len(scripts) - 1):
+    base_script = scripts[i]
+    for j in range(i + 1, len(scripts)):
+      target_script = scripts[j]
+      result[(base_script, target_script)] = compare_interscript(
```
jamesgk: FYI, don't need parentheses between brackets here. Though one may prefer this syntax.

nototools/noto_fonts.py
```diff
@@ -272,6 +272,10 @@ def script_key_to_scripts(script_key):
     return frozenset(['Arab'])
   elif script_key == 'HST':
     raise ValueError('!do not know scripts for HST script key')
+  elif script_key == 'MONO':
+    # Mono doesn't actually support all of Latn, we need a better way to
+    # deal with pseud-script codes like this one.
```
jamesgk: Not sure how much of a difference it makes, but maybe add a TODO here?

jamesgk:
LGTM

nototools/compare_cmap_data.py
dougfelt: I guess the standard injunction against unnecessary parentheses applies.  I'll try to get used to it.

nototools/noto_fonts.py
dougfelt: OK.  This is actually a major TODO that I've made attempts at a few times.  It's not easy getting the semantics right, and these codes are used all over the place.

PR #235, opened 2016-05-31
Update emoji ordering to handle slight changes in data format.

dougfelt:
The csv version of the emoji ordering data file changed a bit, so try
to be a bit more generic in what we handle.

### DIFF
--- a/nototools/generate_emoji_name_data.py
+++ b/nototools/generate_emoji_name_data.py
@@ -96,7 +96,8 @@ def _is_keycap_sequence(cps):
 _KEYCAP_NAMES = {cp: unicode_data.name(cp)[6:] for cp in range(0x30, 0x30+10)}
 
 def _keycap_sequence_name(cps):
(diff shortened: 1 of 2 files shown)


nototools/generate_emoji_name_data.py
```diff
@@ -338,8 +348,9 @@ def main():
       action='store_true')
   args = parser.parse_args()
   generate_names(
-      args.srcdir, args.outfile, args.emoji_ordering, force=args.force,
-      pretty_print=args.pretty_print, verbose=args.verbose)
+      args.srcdir, args.outfile, args.emoji_ordering,
+      eext=args.emoji_ordering_ext, esep = args.emoji_ordering_sep,
```
jamesgk: Minor, but shouldn't have spaces around `=`

jamesgk:
LGTM

nototools/generate_emoji_name_data.py
dougfelt: Oops.

PR #236, opened 2016-05-31
Let defined_characters take a string or a float.

dougfelt:
Passing a string instead of a float caused no errors but returned bad
(but not obviously bad) results.  This seemed as reasonable as raising
an error would have.

### DIFF
--- a/nototools/unicode_data.py
+++ b/nototools/unicode_data.py
@@ -327,6 +327,10 @@ _DEFINED_CHARACTERS_CACHE = {}
 def defined_characters(version=None, scr=None):
     """Returns the set of all defined characters in the Unicode Standard."""
     load_data()
+    # handle common error where version is passed as string, the age test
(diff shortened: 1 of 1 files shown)


nototools/unicode_data.py
```diff
@@ -327,6 +327,9 @@ def create_script_to_chars():
 def defined_characters(version=None, scr=None):
     """Returns the set of all defined characters in the Unicode Standard."""
     load_data()
+    # handle common error where version is passed as string, the age test
+    # will always pass
+    version = float(version)
```
jamesgk: What if version is None?

dougfelt: Pfffft!

There's no code change so small that it doesn't require code review... at least in my case.

dougfelt:
PTAL

jamesgk:
LGTM

PR #237, opened 2016-06-01
Emoji data update

dougfelt:


nototools/unicode_data.py
```diff
@@ -688,6 +693,117 @@ def _load_emoji_data():
...
+            if not line or line[0] == '#':
+                continue
+            m = line_re.match(line)
+            if not m:
+                raise ValueError('Did not match "%s"' % line)
+            seq = tuple(int(s, 16) for s in m.group(1).split())
+            age = float(m.group(2))
+            seq_name = m.group(3)
+            # patch in preferred name for this anomalous value
+            if seq_name == None:
```
jamesgk: Style guide prefers comparing agains None with `is` instead of `==` (not sure why though)
http://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations,
also applies to two later comparisons a few lines down.

nototools/unicode_data.py
```diff
@@ -874,7 +988,43 @@ def codeset(cpname):
...
+      print '+'.join('%04x' % cp for cp in seq), name(seq[0])
+  print 'zwj sequences'
+  for seq, seq_name in sorted(get_emoji_zwj_sequences(age=8.0).items()):
+      print '+'.join('%04x' % cp for cp in seq), seq_name
+
+
+def _dump_emoji_presentation():
+  """Dump presentation info, for testing."""
+
+  ok = 0
```
jamesgk: `ok` is unused

nototools/unicode_data.py
```diff
@@ -874,7 +988,43 @@ def codeset(cpname):
...
+    if cp in get_presentation_default_emoji():
+      presentation = 'emoji'
+      emoji_p += 1
+    elif cp in get_presentation_default_text():
+      presentation = 'text'
+      text_p += 1
+    else:
+      presentation = '<error>'
+    print '%s%04x %5s %s' % (
+        ' ' if cp < 10000 else '', cp, presentation, cp_name)
```
jamesgk: I think this should be `cp < 0x10000`

jamesgk:
LGTM with comments

nototools/unicode_data.py
dougfelt: Um, yeah...

PR #238, opened 2016-06-02
Update cmap requirements for phase 3.

dougfelt:
Two characters formerly designated as emoji were reclassified as non-emoji.
This is indicated in the most recent emoji data file.

Issue #689 in noto-fonts indicates that some power symbols belong together,
which we didn't know before.  These were being placed into different fonts
(4 in Symbols and 1 in Symbols2).  This change puts all 5 characters
into Symbols.  Code was added to noto_cmap_reqs to make this change.

The noto phase3 cmap data was regenerated and reflects these changes.


jamesgk:
LGTM

PR #241, opened 2016-06-15
Generalize ligature unit test

jamesgk:
Before, this tested some Roboto-specific behavior. Now it's
generalized a bit.

### DIFF
--- a/nototools/unittests/font_tests.py
+++ b/nototools/unittests/font_tests.py
@@ -330,25 +330,55 @@ class TestCharacterCoverage(FontTest):
 
 
 class TestLigatures(FontTest):
-    """Tests formation or lack of formation of ligatures."""
+    """Tests formation or lack of formation of ligatures.
+
+    Must have the following attributes:
+        active: List of (parameters, (sequence, sequence, ...)) tuples, for
+            which each sequence should form a ligature when shaped with the
+            given parameters. The ligature should not be formed by default if
+            parameters are provided.
+        inactive: Similar list of tuples, for sequences which should form a
+            ligature under inverse circumstances.
+    """
 
     def setUp(self):
         self.fontfiles, _ = self.loaded_fonts
 
-    def test_lack_of_ff_ligature(self):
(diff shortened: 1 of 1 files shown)


nototools/unittests/font_tests.py
```diff
@@ -330,25 +330,55 @@ def test_format_4_cmap_preset(self):
...
-    """Tests formation or lack of formation of ligatures."""
+    """Tests formation or lack of formation of ligatures.
+
+    Must have the following attributes:
+        active: List of (parameters, (sequence, sequence, ...)) tuples, for
+            which each sequence should form a ligature when shaped with the
+            given parameters. The ligature should not be formed by default if
+            parameters are provided.
+        inactive: Similar list of tuples, for sequences which should form a
+            ligature under inverse circumstances.
```
dougfelt: I'm not quite sure I get what 'inverse circumstances are' without it being spelled out.  It sounds like, if parameters are provided, a ligature should form with all sets of parameters except those provided (including the default case), and if parameters are not provided, a ligature should form with all sets of parameters except the default case.

dougfelt:
LGTM

nototools/unittests/font_tests.py
jamesgk: Yea, I'm certainly not using "inverse" in a strictly correct sense here. Basically to spell it out I would say:

```
        inactive: List of (parameters, (sequence, sequence, ...)) tuples, for
            which each sequence should not form a ligature when shaped with the
            given parameters. The ligature should be formed by default if
            parameters are provided.
```

but I didn't want to essentially repeat the last attribute description with just the word "not" moved. What if it said "opposite" or "the reverse" circumstances? Or, any other suggestions?

dougfelt: Given the same arguments, the inactive case fails where the active case succeeds, and vice-versa.  I think you just say something like that.  That really is what is going on-- there's no 'inverse' or 'reverse' circumstances.

PR #247, opened 2016-06-22
[unittests] Simplify ligature test

jamesgk:
Not expecting ligatures to be inactive when they're active with
parameters gives more flexibility. One might want to check that a
ligature works both with and without certain parameters.


dougfelt:
LGTM

PR #250, opened 2016-06-27
Build fix

### LINKED ISSUE #135: breaks with a narrow python build
$ python add_svg_glyphs.py
Traceback (most recent call last):
  File "add_svg_glyphs.py", line 37, in <module>
    import add_emoji_gsub
  File "/Users/gilbahat/fonts/noto-fonts-master/add_emoji_gsub.py", line 28, in <module>
    from nototools import font_data
  File "/Users/gilbahat/fonts/noto-fonts-master/nototools/font_data.py", line 21, in <module>
    from nototools import opentype_data
  File "/Users/gilbahat/fonts/noto-fonts-master/nototools/opentype_data.py", line 330, in <module>
    _set_ompl()
  File "/Users/gilbahat/fonts/noto-fonts-master/nototools/opentype_data.py", line 33, in _set_ompl
    unicode_data.load_data()
  File "/Users/gilbahat/fonts/noto-fonts-master/nototools/unicode_data.py", line 75, in load_data
    _load_unicode_data_txt()
  File "/Users/gilbahat/fonts/noto-fonts-master/nototools/unicode_data.py", line 457, in _load_unicode_data_txt
    decomposition = [unichr(int(char, 16)) for char in decomposition]
ValueError: unichr() arg not in range(0x10000) (narrow Python build)

a narrow python build will be a common situation... and specifically there's this: https://github.com/Homebrew/homebrew/pull/32368 :(

adrientetar:
Fixes #135.

r? @jamesgk

### DIFF
--- a/nototools/unittests/font_tests.py
+++ b/nototools/unittests/font_tests.py
@@ -26,7 +26,7 @@ from nototools import coverage
 from nototools import font_data
 from nototools import noto_fonts
 from nototools import unicode_data
-from nototools.pens.glyph_area_pen import GlyphAreaPen
+from nototools.glyph_area_pen import GlyphAreaPen
 from nototools.unittests import layout
 
 
@@ -57,7 +57,7 @@ def load_fonts(patterns, expected_count=None, font_class=None):
         all_font_files += glob.glob(pattern)
     all_fonts = [font_class(font) for font in all_font_files]
     if expected_count:
-        assert len(all_font_files) == expected_count
+        assert len(all_font_files) == expected_count, "got %d fonts, expected %d." % (len(all_font_files), expected_count)
     return all_font_files, all_fonts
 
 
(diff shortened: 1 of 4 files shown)


jamesgk:
Let's also delete glyphAreaPen from this repo. The import in nototools/shape_diff.py also needs to be updated.

adrientetar:
Updated.

nototools/unittests/font_tests.py
```diff
@@ -592,9 +592,9 @@ def setUp(self):
             if name in self.whitelist:
                 continue
             master_a[name].draw(pen_a)
-            area_a = pen_a.unload()
+            area_a = pen_a.value
```
jamesgk: Oh, I didn't think about this. I think we need to actually create a new pen each time we draw a glyph (or add the unload method to the fonttools pen), otherwise the areas compound on one another.

anthrotype: or just set `pen.value = 0`

jamesgk: Yea, that's better than making a bunch of new pens. I still prefer an unload method but don't have strong feelings.

jamesgk: Based on https://github.com/behdad/fonttools/pull/627#issuecomment-228894659 we need to reset the pen value after it's used here (and also change the calls to unload in shape_diff).

jamesgk: (Ah, discussion is still ongoing, disregard for now)

adrientetar:
Updated.

PR #257, opened 2016-07-06
Coverage data

dougfelt:
Adds two tools for generating cmap data from one or more fonts, and reporting on these data.  Also adds a utility function that generates a list of paths from a list of directories and/or list of files or a file containing a list of paths.

### DIFF
--- a/nototools/generate_coverage_data.py
+++ b/nototools/generate_coverage_data.py
@@ -1,0 +1,245 @@
+#!/usr/bin/python
+#
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import collections
+import datetime
+from os import path
+import sys
+
+from nototools import cmap_data
+from nototools import coverage
+from nototools import tool_utils
+
+from xml.etree import ElementTree as ET
+
+"""Generate a coverage.xml file listing the codepoints covered by
+a font family plus a name."""
+
+default_coverage_file = 'noto_cmap_phase2.xml'
+
+MetaData = collections.namedtuple('MetaData', 'date,program,args')
+CmapData = collections.namedtuple('CmapData', 'name,ranges')
+CoverageData = collections.namedtuple('CoverageData', 'metadata,cmapdata')
+
+
+def get_cps_from_files(paths):
+  """Return a tuple of the cps and the paths that we actually read."""
+  cps = set()
+  new_paths = set()
+  for f in paths:
+    ext = path.splitext(f)[1]
+    if ext not in ['.otf', '.ttf']:
+      continue
+    cps |= coverage.character_set(f)
+    new_paths.add(f)
+  return cps, sorted(new_paths)
+
+
+def get_cps_from_cmap_data_file(data_file):
+  cps = set()
+  data = cmap_data.read_cmap_data_file(data_file)
+  for row in data.table.rows:
+    cps |= tool_utils.parse_int_ranges(row.ranges)
+  return cps
+
+
+def _create_metadata(**kwargs):
+  """Create a MetaData object from the args.  'date' defaults to today's
+  date."""
+  date = str(kwargs.pop('date', datetime.date.today()))
+  program = str(kwargs.pop('program', 'generate_coverage_data'))
+  arglist = [
+      (k, v) for k, v in sorted(kwargs.iteritems())
+      if v is not None]
+  return MetaData(date, program, arglist)
+
+
+def _create_cmapdata(name, cmap):
+  """Create a CmapData object from the name and cmap."""
+  return CmapData(name, tool_utils.write_int_ranges(cmap))
+
+
+def create(name, cps, paths=None, cmap_data=None):
+  """Generate the coverage data object."""
+  metadata = _create_metadata(paths=paths, cmap_data=cmap_data)
+  cmapdata = _create_cmapdata(name, cps)
+  return CoverageData(metadata, cmapdata)
+
+
+def _common_path_prefix(items):
+  """Assuming items is an array of paths using path.sep as a path separator,
+  return a common path prefix of the items."""
+  prefix = None
+  if len(items) <= 1:
+    return ''
+  for item in items:
+    if prefix is None:
+      last = item.rfind(path.sep)
+      if last < 0:
+        return ''
+      prefix = item[:last + 1]
+      continue
+    end = len(item)
+    while True:
+      last = item.rfind(path.sep, 0, end)
+      if last < 0:
+        return ''
+      item_prefix = item[:last + 1]
+      if prefix == item_prefix:
+        break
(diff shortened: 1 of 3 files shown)


nototools/tool_utils.py
```diff
@@ -358,3 +358,41 @@ def setup_logging(loglevel, quiet_ttx=True):
...
+# see noto_fonts.NOTO_FONT_PATHS
+NOTO_FONT_PATHS = [
+    '[fonts]/hinted', '[fonts]/unhinted', '[emoji]/fonts', '[cjk]']
+
+
+def collect_paths(dirs, files):
+  """Return a collection of all files in any of the listed dirs, and
+  the listed files.  Can use noto short paths.  A file name starting
+  with '@' is interpreted as the name of a file containing a list
+  of filenames one per line.  The short name '[noto] refers to
```
jamesgk: Missing a single quote after [noto]?

nototools/generate_coverage_data.py
```diff
@@ -0,0 +1,253 @@
...
+
+def create(name, cps, paths=None, cmap_data=None):
+  """Generate the coverage data object."""
+  metadata = _create_metadata(paths=paths, cmap_data=cmap_data)
+  cmapdata = _create_cmapdata(name, cps)
+  return CoverageData(metadata, cmapdata)
+
+
+def _common_path_prefix(items):
+  """Assuming items is an array of paths using '/' as a path separator, return a
```
jamesgk: Could also use os.path.sep

nototools/generate_coverage_data.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _build_meta_elem(metadata):
+  """Convert metadata to ET node."""
+  meta = ET.Element('meta', date=metadata.date, program=metadata.program)
+  if metadata.args:
+    args = ET.Element('args')
+    for k, v in metadata.args:
+      if isinstance(v, list):
+        arg = ET.Element(k, {'isList': 'true'})
```
jamesgk: Is there any difference between providing attributes as keyword args vs. a dictionary? (if not, why use both)

nototools/generate_coverage_data.py
```diff
@@ -0,0 +1,253 @@
...
+      args.append(arg)
+    meta.append(args)
+  return meta
+
+
+def _build_cmap_elem(cmapdata):
+  """Convert data to ET node."""
+  data = ET.Element('data')
+  name = ET.Element('name')
+  name.text=cmapdata.name
```
jamesgk: Whitespace

nototools/generate_coverage_data.py
```diff
@@ -0,0 +1,253 @@
...
+  coverage = create(args.name, cps, paths=paths, cmap_data=args.cmap_data)
+  write(coverage, args.output_file)
+
+  # test
+  if args.output_file:
+    new_coverage = read(args.output_file)
+    if new_coverage == coverage:
+      print 'OK'
+    else:
+      print 'FAIL'
```
jamesgk: Did you mean to keep this test in? If so, suggesting to make this message more informative.

nototools/report_coverage_data.py
```diff
@@ -0,0 +1,286 @@
...
+  if exclude_ranges:
+    defined_cps -= tool_utils.parse_int_ranges(exclude_ranges)
+  return defined_cps
+
+
+def get_coverages(coverage_files):
+  return [generate_coverage_data.read(f) for f in coverage_files]
+
+
+def get_covered_cps_from_files(filepaths):
```
jamesgk: This function is unused; not sure if you meant to keep it in.

nototools/report_coverage_data.py
```diff
@@ -0,0 +1,286 @@
...
+    tot_ht = 1 + int(math.floor(float(num_in_block) / cp_limit))
+    tot_px = int(float(chart_width) * num_in_block / max_block_size / tot_ht)
+    vir_cp_limit = int(float(num_in_block) / tot_ht)
+    out_file.write('    <tr>')
+    out_file.write('<td class="range">')
+    out_file.write(range_str)
+    out_file.write('<td class="name">')
+    out_file.write(name)
+    out_file.write('<td class="count">')
+    out_file.write(str(num_in_block))
```
jamesgk: You could do one call to write with a single formatted string here, might be cleaner.

nototools/report_coverage_data.py
```diff
@@ -0,0 +1,286 @@
...
+      num_covered = len(block_covered_cps)
+      pct = '%d%%' % int(100.0 * num_covered / num_in_block)
+      bar_ht = min(
+          tot_ht, 1 + int(math.floor(float(num_covered) / vir_cp_limit)))
+      bar_px = int(float(chart_width) * num_covered / max_block_size / bar_ht)
+      out_file.write('<td class="count">')
+      out_file.write(str(num_covered))
+      out_file.write('<td class="pct">')
+      out_file.write(pct)
+      out_file.write('<td><div class="chart">')
```
jamesgk: Same comment as above. Though I guess with the stuff right below here it's a little hairy either way.

nototools/report_coverage_data.py
```diff
@@ -0,0 +1,286 @@
...
+      out_file.write('</div></div></div>\n')
+  out_file.write(FOOTER)
+
+
+def write_block_coverage_text(block_data, names, msg, out_file=sys.stdout):
+  block_data.sort()
+  print >> out_file, msg
+  name_len = max(len(t[2]) for t in block_data)
+  fmt_str = '%%%ds' % name_len
+  fmt_str = '%13s ' + fmt_str + ' %5s'
```
jamesgk: Could combine these two lines?

jamesgk:
LGTM

nototools/tool_utils.py
dougfelt: Yes.

nototools/generate_coverage_data.py
dougfelt: Yes.

nototools/generate_coverage_data.py
dougfelt: Yes, providing them as args is simpler.

nototools/generate_coverage_data.py
dougfelt: Yes.

nototools/generate_coverage_data.py
dougfelt: No, meant to get rid of that.

nototools/report_coverage_data.py
dougfelt: No, meant to remove that.

nototools/report_coverage_data.py
dougfelt: Yeah, well, I kind of prefer to write each table cell separately, the format is a bit easier to follow that way, at least it seems to me.  And then I would end up formatting strings with one item and then writing the formatted string to the file... When I can just write to the file directly it seems kind of either/or.  I'll change it to write each cell on one line rather than alternating element then data then element then data...

nototools/report_coverage_data.py
dougfelt: Modified to write elem + value on single lines.

nototools/report_coverage_data.py
jamesgk: Oh yea, I think that's a good compromise.

nototools/report_coverage_data.py
dougfelt: I could, I guess.  I'd have to parenthesize the first line inside the second expression or else I think the string concatenation would take precedence.  I don't think python supports formatting with field-width arguments like C printf, so it's a bit messy no matter what I do.  I'll just leave it.

jamesgk: I guess I was thinking: `fmt_str = '%%13s %%%ds %%5s' % name_len`. But that does look a bit odd.

PR #259, opened 2016-07-13
Update tool to allow specification of additional codepoints.

dougfelt:
This gets called from the emoji makefile.  Occasionally we build fonts
without the supporting data present, and this lets us hack in additional
codepoints to build these tests.

### DIFF
--- a/nototools/add_vs_cmap.py
+++ b/nototools/add_vs_cmap.py
@@ -35,6 +35,7 @@ from fontTools.ttLib.tables import _c_m_a_p
 
 from nototools import font_data
 from nototools import unicode_data
+from nototools import tool_utils
 
 VS_TEXT = 0xFE0E
 VS_EMOJI = 0xFE0F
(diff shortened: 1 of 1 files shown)


nototools/add_vs_cmap.py
```diff
@@ -56,7 +57,7 @@ def modify_font(font_name, font, presentation, emoji_variants):
 
 
 def modify_fonts(font_names, presentation='emoji', output=None, suffix=None,
-                 dst_dir=None):
+                 dst_dir=None, vs_added=None):
```
jamesgk: What does vs stand for? Suggest maybe changing this name.

jamesgk:
LGTM

nototools/add_vs_cmap.py
dougfelt: variation selector.

this is how the entire tool is named, are you suggesting to change that too?

jamesgk: Ah, didn't notice that. In context it makes more sense, never mind.

PR #264, opened 2016-07-15
Update emoji data to version 4.

dougfelt:
This includes the gendered sequences.

The emoji data isn't quite enough, we also need to know that some additional
characters take an emoji variation selector-- this got patched into the
code that parses the data, since StandardVariants.txt doesn't have it yet.

The emoji data format changed a bit-- names are now formally a field, and
all sequences now have 'names', so we parse those names.  Two APIs have
changed to return dicts instead of sets as a result.


nototools/unicode_data.py
```diff
@@ -849,6 +837,11 @@ def _load_unicode_emoji_variants():
       m = line_re.match(line)
       if m:
         emoji_variants.add(int(m.group(1), 16))
+
+  # temporarily hard-code in some post-9.0 proposed excerptions used
```
jamesgk: *exceptions

jamesgk:
LGTM

PR #265, opened 2016-07-18
Fix notodiff on OSX

jamesgk:
Fixes https://github.com/googlei18n/nototools/issues/254
Fixes https://github.com/googlei18n/nototools/issues/255


dougfelt:
LGTM

PR #266, opened 2016-07-19
[notodiff] Don't render if one glyph is zero-width

jamesgk:
If one glyph in a comparison will be rendered with width zero (either
because its width is zero or because it has class mark), the rendered
comparison is not very helpful. This prints an error message instead.

The goal is to resolve https://github.com/googlei18n/fontmake/issues/105


dougfelt:
LGTM

PR #278, opened 2016-08-22
Generate html to compare dingbats fonts with noto

dougfelt:
Generate an html file that generates a table comparing a target font
against multiple reference fonts.  Supports multiple target fonts.
Characters from each target font are in a specified order, matching
characters from reference fonts are displayed on the same line together
with some basic unicode info.


jamesgk:
LGTM

PR #279, opened 2016-08-30
Add NamesList.txt unicode data and apis to read it.

dougfelt:
NamesList.txt is used for generating the unicode book.  The format is
designed for that purpose and not really as a data format.  However, this
is the only way to access information about related characters (that are
not equivalent) and also other names that some characters are known by
(e.g. LINE FEED).  This provides apis to get these two pieces of information
about a character, but does not attempt to parse the entirety of the
information from the NamesList.txt file.


roozbehp:
Generally, parsing NamesLists.txt is not recommended by its editors. I understand the related characters data only exists there (I'm curious how you want to use it), but the alternative names is better read from http://www.unicode.org/Public/UCD/latest/ucd/NameAliases.txt

dougfelt:
I was interested in seeing if I could identify clusters of characters to compare (primarily symbols characters) by looking at the related characters data and names data.  Since I was looking at that file anyway I thought I could get alternative names from there.  Are there differences between this data and that in NameAliases.txt?

roozbehp:
I don't know if there definitely are differences, unfortunately, but I assume there may be or will be in the future. But I know that NamesList.txt is only intended to generate the glyph charts, and not for consumption by tools. So I highly recommend using NameAliases.txt which is the canonical location of alternative names. The version of alternative names in NamesLists.txt is just there so people who happen to look at the charts have more information at hand.

But finding clusters of characters to compare is a legitimate use of NamesLists.txt, since the cross references data only exists publicly in NamesLists.txt.

jamesgk:
Code LGTM

dougfelt:
@jamesgk, PTA(nother)L

jamesgk:
LGTM

PR #280, opened 2016-08-30
Dingbats html

dougfelt:
Further modifications to the glyph comparison html tool.

Main changes are:
1) refactor targets/references so that all font descriptions are the same format
2) separate definition of character set from a reference font, to make it possible to generate tables comparing arbitrary collections of characters
3) implement code that can define/read character set data from files or inline
4) output formatting tweaks

For more information see the previous commits.

### DIFF
--- a/nototools/generate_dingbats_html.py
+++ b/nototools/generate_dingbats_html.py
@@ -81,6 +81,16 @@ class CodeList(object):
   points that the font knows how to display."""
 
   @staticmethod
+  def fromfile(filename):
+    if filename.endswith('_codes.txt'):
+      return CodeList.frompairfile(filename)
+    elif filename.endswith('_cmap.txt'):
+      return CodeList.fromrangefile(filename)
+    else:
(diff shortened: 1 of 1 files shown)


nototools/generate_dingbats_html.py
```diff
@@ -320,54 +505,61 @@ def _generate_styles(targets, references):
...
+    cps = unichr(codelist.mapped_code(cp))
+    return (context % cps) if context else cps
 
   lines = ['<h3 id="target_%d">%s</h3>' % (index, name)]
   lines.append('<table>')
-  header = _generate_header(target, references)
+  header = _generate_header(used_fonts)
   linecount = 0
-  for codepair in codes:
+  # cp0 is the valid unicode, cp1 is the PUA if the font uses PUA encodings
```
jamesgk: cp1 is no longer used

jamesgk:
LGTM

PR #287, opened 2016-09-14
Read dingbats char compare data from file.

dougfelt:


nototools/generate_dingbats_html.py
```diff
@@ -385,47 +385,111 @@ def _load_fonts_and_targets(font_data, target_data, data_dir):
...
+      if not rest:
+        rest = None
+      if name == 'prefer':
+        prefer_fallback = rest
+      elif name == 'omit':
+        omit_fallback = rest
+      else:
+        raise Exception('default only understands \'prefer\' or \'omit\'')
+      continue
+    info = line.split(';')
```
jungshik: Would you summarize the line format here again (2 ~ 4 fields; what each field is for, etc)?

dougfelt: - first field is a name for the character list
- second field is a specification of the character list, basically either a list of code points and ranges, or a list of code points in specified order, or the name of a file that lists code points/alternates in specified order
- third field is the ids of the 'preferred fonts', this forces these columns to be at the rightmost of the table, i.e. in 'reference position'.  For example if the set of codepoints comes from wingdings then we want wingdings to be the rightmost column.
- fourth field is the ids of the 'excluded fonts', this forces these fonts to be omitted from the table.  normally we provide a column for each font that has at least one codepoint in the set, this lets us exclude some that we're not interested in.

dougfelt:
Well, not yet.  The data I currently have includes fonts that I can't push,
and they and the files of codepoint data and the font and target
specification files are all in a single directory that you target the tool
at, for convenience.  It would be easier to understand the code if there
were examples to look at.  I don't have a good sense of what makes sense to
preserve in the repo, though.  Currently the data is just stuff I use to
generate the html tables comparing some noto fonts and the repertoire of
each of the dingbats fonts (and now arbitrary lists of codepoints as
displayed by the data files).

nototools/generate_dingbats_html.py
jungshik: Sorry that I was not clear. Thank you for the explanation. 

I read your function-level comment and know what each field is. (but I like your explanation above  :-) ).  What I meant was  adding in-place comments would make it easier for other readers to understand.

nototools/generate_dingbats_html.py
```diff
@@ -385,47 +385,111 @@ def _load_fonts_and_targets(font_data, target_data, data_dir):
...
+        items = fallback
+      info[index] = items
+
+  prefer_fallback = None
+  omit_fallback = None
+  defines = {}
+  target_data = []
+  for line in strip_comments_from_file(filename):
+    if line.startswith('!define '):
+      name, rest = line[8:].split('=')
```
jungshik: nit: forgot to mention earlier.   optional. I wonder if the following would be better. 
```
kDefineDirective = '!define '  
if line.startswith(kDefineDirective):
  name, rest = line[len(kDefineDirective)].split('=')
```

The same can be said of  '!default '.

PR #288, opened 2016-09-19
Added third_party/ as package data.

n7s:
Just putting the data that notodiff requires in the right folder (like @jamesgk showed me at AtypI.)


n7s:
I signed it

jamesgk:
Thanks!

PR #289, opened 2016-09-28
Update Symbols/Symbols2 data and tools used to review it

dougfelt:
The primary driver of these changes was the review of the proposed Symbols2 font glyphs.

This prompted a few things:
1) the addition of some unicode data (to specify additional emoji that should get text presentation forms and/or B/W glyphs).
2) updates to the tool that determines which fonts support which codepoints, to better allocate codepoints between Symbols (multiple weight masters) and Symbols2 (single weight master)
3) updates to the tool that generates an html/text report comparing the same codepoints in multiple fonts.


nototools/generate_dingbats_html.py
```diff
@@ -512,11 +512,19 @@ def add_index_list_or_defined(info, index, fallback, defines):
...
-      unicode_data.get_emoji() - unicode_data.get_unicode_emoji_variants(
-          'proposed_extra'))
-
+def _flagged_name(cp, flag_sets):
+  """Prepend any flags to cp's unicode name, and return."""
+  name = unicode_data.name(cp)
+  flags = []
+  for k, v in sorted(flag_sets.iteritems()):
+    if (cp in v[0]) == v[1]:
+      flags.append(k)
```
jungshik: How about repeating this comment taken from  |flag_sets| here? 
"# True means set flag if cp in set, False means set if not in set"

nototools/noto_cmap_reqs.py
```diff
@@ -435,7 +451,7 @@ def _reassign_common_by_block(cmap_ops):
     'Enclosed Alphanumerics': 'Zsym',
     'Box Drawing': 'MONO',
     'Block Elements': 'MONO',
-    'Geometric Shapes': 'Zsym',
+    'Geometric Shapes': 'SYM2', # change
```
jungshik: Just wondering (not proposing any change): Can you think of any scenario in which using 'SYM2' (not valid per ISO 15924) will bite us in the future  ? Extremely unlikely because it's just our internal code and 'SYM2' is infinitely more "readable" than "Qaaa' (picked from the private use range).

nototools/generate_dingbats_html.py
dougfelt: Sure.

nototools/noto_cmap_reqs.py
dougfelt: Yes, these all are internal codes that just happen to look a lot like script codes.  I've thought about trying to rename them so that it is clear they are not really script codes, but there's stuff all over the place that assumes that the string is a script code except for a few exceptions, for example to get the English name of the 'script' from the cldr data...  To make things worse, these are also part of a font family code, which looks a lot like a bcp47 locale string, except it's not and has some extra stuff in it, again in only a few exceptional cases.  Anyway, yes, you're right to wonder about this, and I've made attempts in the past to change it, but haven't managed to because in the end the existing mostly-script-codes-plus-a-few-fake-codes seem to work well enough and I have to change all kinds of code to clean this up.  I think we just remain careful.  I use all caps for the fake codes and don't always restrict to 4 letters so collisions should be unlikely  (though BCP47 isn't case-sensitive so it's still possible).

PR #290, opened 2016-10-04
[unittests] Add test for glyph bounds

jamesgk:
This allows testing that certain glyphs do or do not extend beyond
certain bounds. Complicating the process slightly is that I wanted
to be able to specify different bounds for different weights as a
product of the weight differences and some user-provided multiplier.

### DIFF
--- a/nototools/unittests/font_tests.py
+++ b/nototools/unittests/font_tests.py
@@ -570,8 +570,9 @@ class TestGlyphBounds(FontTest):
      - names: a list containing glyph names as strings.
      - bounds: a list (xmin, ymin, xmax, ymax), values can be None.
      - multiplier: a value which, when multiplied by a font's weight class
-         value's difference from 400 (regular), gives a bound adjustment for
-         that font.
+         value minus 400 (regular), gives a bound adjustment for that font.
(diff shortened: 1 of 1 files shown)


nototools/unittests/font_tests.py
```diff
@@ -561,6 +561,59 @@ def test_height_of_lowercase_o(self):
...
+    """Tests that certain glyphs do or don't go outside of given bounds.
+
+    Glyphs must be designated with two class attribute lists "should_exceed" and
+    "should_not_exceed", both containing tuples each with the following four
+    elements:
+     - names: a list containing glyph names as strings.
+     - bounds: a list (xmin, ymin, xmax, ymax), values can be None.
+     - multiplier: a value which, when multiplied by a font's weight class
+         value's difference from 400 (regular), gives a bound adjustment for
+         that font.
```
dougfelt: positive values adjust the bounds rectangle outward? (xmin moves left, xmax moves right...)

dougfelt:
LGTM

nototools/unittests/font_tests.py
jamesgk: Right. I'll add a note to the documentation.

PR #291, opened 2016-10-05
Update cmap data for phase 3

dougfelt:


nototools/noto_cmap_reqs.py
```diff
@@ -789,6 +789,19 @@ def _assign_nastaliq(cmap_ops):
   """)
   cmap_ops.add_all(additional_arabic, 'Aran')
 
+  # noto-fonts#597 requests exclamation point
+  # noto-fonts#449 requests european digits
+  european_digits = tool_utils.parse_int_ranges('0021 0030-0039')
+  cmap_ops.add_all(european_digits, 'Aran')
+
+  # noto-fonts#368 requests these characters
+  extra_arabic_1 = tool_utils.parse_int_ranges('067b  0684 068a 06b3 0759 0768')
```
jungshik: These characters were  in noto_cmap_phase3.xml before this CL. So, the above line is just to sync this script with xml, I guess?

dougfelt: Mostly just to document the source of the request.

PR #294, opened 2016-10-08
Comp tool flags

dougfelt:
Most of this is a change to the tool to be still more driven by data files.  There's also a tweak to change the width of the output table.

### DIFF
--- a/nototools/generate_dingbats_html.py
+++ b/nototools/generate_dingbats_html.py
@@ -47,7 +47,8 @@ _HTML_HEADER_TEMPLATE = """<!DOCTYPE html>
     tr.head { font-weight: bold; font-size: 12pt;
               border-style: solid; border-width: 1px; border-color: black;
               border-collapse: separate }
-    .code, .age, .name { font-size: 12pt; text-align: left }
+    .code, .age { font-size: 12pt; text-align: left }
+    .name { font-size: 10pt; text-align:left }
     .key { background-color: white; font-size: 12pt; border-collapse: separate;
            margin-top: 0; border-spacing: 10px 0; text-align: left }
   </style>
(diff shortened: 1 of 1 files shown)


nototools/generate_dingbats_html.py
```diff
@@ -667,6 +679,7 @@ def context_string(codelist, cp):
         line.append('<td>&nbsp;')
     line.append('<td class="age">%s' % unicode_data.age(cp))
     name = _flagged_name(cp, flag_sets)
+    name = _break_name(name)
     line.append('<td class="name">%s' % name)
```
jungshik: FYI: You can use n-th-child selector to avoid adding ```class="<class name>"``` for every cell. 
See https://developer.mozilla.org/ko/docs/Web/CSS/:nth-child

nototools/generate_dingbats_html.py
```diff
@@ -667,6 +679,7 @@ def context_string(codelist, cp):
         line.append('<td>&nbsp;')
     line.append('<td class="age">%s' % unicode_data.age(cp))
     name = _flagged_name(cp, flag_sets)
+    name = _break_name(name)
```
jungshik: An alternative would be to set the column width and let browsers wrap lines.

nototools/generate_dingbats_html.py
```diff
@@ -395,13 +395,87 @@ def _load_targets(target_data, fonts, data_dir, codelist_map):
...
+      if not result:
+        # it appers that python 'optimizes' by replacing the lhs by rhs if lhs
+        # is an empty set, but this changes the type of lhs to frozenset...
+        result = set(codes)
+      else:
+        result |= codes
+    elif op == '&':
+      result &= codes
+    else:
+      result -= codes
```
jungshik: Shouldn't you need a guard against an unsupported op (op other than {|, &, -} )? Or, you're fine with  treating anything other than {&, |} as '-'?

nototools/generate_dingbats_html.py
dougfelt: Thanks for the suggesstion.  Each table has a different number of columns, so I'd have to generate this data for each table separately, no?.  The first column and the last two columns can be dealt with this way, so I'll do that.

nototools/generate_dingbats_html.py
dougfelt: The scanner ensures that these are the only three operation characters that might be encountered here, but it is more robust to check here as well, true.

nototools/generate_dingbats_html.py
dougfelt: Yeah, this was a hack.  I'll just do the right thing.

dougfelt:
@jungshik, PTAL

PR #297, opened 2016-10-18
[notodiff] Fix shape ratio calculation

jamesgk:
Before, this didn't account for negative areas.

Should fix https://github.com/googlei18n/nototools/issues/296


dougfelt:
LGTM

PR #298, opened 2016-10-19
[hb_input] Try to make input more consistent

jamesgk:
In addition to the notes from the first commit, should also fix https://github.com/googlei18n/nototools/issues/283.


dougfelt:
LGTM

PR #299, opened 2016-10-20
Add a script to generate an html page from our sample texts.

dougfelt:


dougfelt:
Marek, this was a request for a code review, and I'd assigned it to Jungshik for that reason.  

The way to handle this is to leave a comment with LGTM if you have read the code and believe it is ok, or leave comments on the code if not.  I would do the actual merge.

marekjez86:
what should I do now? 
LGTM 
:-(

dougfelt:
Nothing, you merged it and there's no point in backing it out.  It's just a
tool so it's no big deal, it's not run by other code.  I just wanted you to
be aware what the process is for internal pull requests involving code.

PR #300, opened 2016-10-26
Shape diff fixes

jamesgk:


jamesgk:
cc @marekjez86, this will swap the colors in notodiff rendering output (red for old, green for new -- now consistent with fontdiff).

PR #307, opened 2016-10-31
Add information about metrics, support char sequences.

dougfelt:
This change does three main things (wasn't able to easily pull them apart,
sorry):

- Adds reporting on glyph metrics (side bearings, advance, glyph vertical
  center) as well as the ability to take a metrics file listing desired
  lsb, adv, and vertical center and show these where they differ from the
  current metrics.

- Adds links showing which groups (other than the current group) each
  character is listed in.

- Adds display of (currently hard-coded) character sequences; this is so
  we can display combining mark behavior.

In addition, this also makes the tool executable, updates some
comments, and specifies (hard-coded) a font to use for the context
chars.


dougfelt:
@jungshik, did you have feedback on this?

PR #309, opened 2016-11-03
notodiff addition and improvement

jamesgk:


dougfelt:
LGTM

BTW, nothing in the file appears to describe the different types of comparison and why you'd choose one over the other.  Is there some other document that describes this?

jamesgk:
I put a little bit of information at the top of shape_diff.py. Maybe that info should be expanded and/or put into notodiff.py.

dougfelt:
Thanks.  Still LGTM...

jamesgk:
Yikes, forgot about this.

PR #311, opened 2016-11-04
Diff threshold parameter added.

iverg:
Diff threshold defines minimal diff value to report. Rendered diff
type can now handle ratio_diffs flag properly. Only PNG files
that differ remain in render directory after render diff operation.

### DIFF
--- a/nototools/notodiff.py
+++ b/nototools/notodiff.py
@@ -30,7 +30,7 @@ import os
 from nototools import gpos_diff, gsub_diff, shape_diff
 
 
-def _shape(path_a, path_b, stats, diff_type, render_path):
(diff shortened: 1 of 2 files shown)


marekjez86:
@dougfelt : could you review it?

jamesgk:
LGTM, except that I think we should combine gpos_bound and diff_threshold into a single argument called diff_threshold.

dougfelt:
Looks like this is also waiting on the CLA?

jamesgk:
@iverg, could you incorporate the changes I suggested?

jamesgk:
I'll just do it. Anyways, I don't think we have to worry about the CLA since Igor's a googlei18n member.

PR #319, opened 2016-11-21
notodiff fixes

jamesgk:
Should fix https://github.com/googlei18n/nototools/issues/315

### DIFF
--- a/nototools/hb_input.py
+++ b/nototools/hb_input.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 
-from fontTools.ttLib import TTFont
+from __future__ import division, print_function
 
+from fontTools.ttLib import TTFont
 from nototools import summary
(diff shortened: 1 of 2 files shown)


nototools/hb_input.py
```diff
@@ -109,7 +121,14 @@ def input_from_name(self, name, features=(), seen=None, pad=False):
         # recursive calls that it makes might have themselves returned None,
         # but we should avoid returning None here if there are other options
         inputs = [i for i in inputs if i is not None]
```
jamesgk: Actually the reason why I avoid using filter is that it behaves differently between python 2 and 3 (where it returns a filter object). For example `bool(filter(None, [0]))` is False in python 2 but True in python 3. Not an issue (yet) for nototools but still something I like to avoid.

PR #327, opened 2016-12-06
notodiff adjustments for gsub and gpos diffing

jamesgk:


nototools/hb_input_test.py
```diff
@@ -135,76 +135,84 @@ class definition end
...
+    table.platEncID = 1
+    table.language = 0
+    table.cmap = {AGL2UV[n]: n for n in glyph_order if n in AGL2UV}
+    cmap.tableVersion = 0
+    cmap.tables = [table]
+
+    font['glyf'] = glyf = newTable('glyf')
+    glyf.glyphs = {}
+    glyf.glyphOrder = glyph_order
+    for name in glyph_order:
```
jamesgk: Yea, the function imported from feaLib creates a test font with a hundred or so dummy glyphs.

PR #341, opened 2016-12-13
Quick tool to generate lists of codepoints from cmap data.

dougfelt:


nototools/gen_cplist.py
```diff
@@ -0,0 +1,110 @@
...
+  if len(scripts) != len(namepats):
+    if len(namepats) != 1:
+      raise ValueError(
+          'Have %d script%s but %d namepats' %
+          (len(scripts), '' if len(scripts) == 1 else 's', len(namepats)))
+    if '%s' not in namepats[0] and len(scripts) > 1:
+      raise ValueError(
+          'Have multiple scripts but single namepat "%s" has no substitution'
+          % namepats[0])
+  if len(namepats) < len(scripts):
```
jamesgk: We don't need this if statement, right? Just put the following line in the preceding block I think.

nototools/gen_cplist.py
```diff
@@ -0,0 +1,110 @@
...
+      '-c', '--cmap_file', help='cmap data file to use (default %s)' %
+      default_cmap, default=default_cmap, metavar='file')
+  parser.add_argument(
+      '-d', '--dest_dir', help='directory for output, (defaults to current '
+      'directory)', metavar='dir', default='.')
+  parser.add_argument(
+      '-s', '--scripts', help='script ids of data to output', nargs='+',
+      metavar='id', required=True)
+  parser.add_argument(
+      '-n', '--namepats', help='name patterns (default "cps_%%s.txt")',
```
jamesgk: Might want to specify what this name pattern is used for here.

nototools/gen_cplist.py
dougfelt: Yeah, I completely missed that, thanks.

nototools/gen_cplist.py
dougfelt: OK, expanded the help string a bit.

PR #343, opened 2016-12-14
Use boundsPen to determine vertical extents in render.py.

dougfelt:
This changes the types of arguments to get_glyph_cleaned_extents.

The previous code attempted to optimize computing the vertical extents
by making assumptions about what kinds of transforms would be applied
to components (translations and positive scales).  This didn't support
flips, and also didn't support rotations, which some phase3 fonts use.
BoundsPen supports the omission of single-point contours, like the old
code did, and it's more general, so rather than try to extend the old
optimizations to cover these new cases it seemed cleanest to just use
BoundsPen.

This however means the old argument types (the 'glyf' table and
_g_l_y_f.Glyph objects returned from it) needed to be converted to
_TTGlyph and TTGlyphSet, and it seems best to instead just pass them
in.

The two direct callers of this api (and the caller in render.py)
were changed to pass in the right types.

This of course breaks anyone outside of nototools who might be using
this API.

### DIFF
--- a/nototools/render.py
+++ b/nototools/render.py
@@ -24,6 +24,7 @@ import subprocess
 
 import font_caching
 
+from fontTools.pens.boundsPen import BoundsPen
 
 def min_with_none(first, second):
     """Returns the minimum of the two inputs, ignoring Nones."""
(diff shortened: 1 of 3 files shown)


nototools/render.py
```diff
@@ -68,23 +78,32 @@ def get_glyph_cleaned_extents(glyph, glyf_table):
...
+            try:
+                for component in glyph.components:
+                    component_ymin, component_ymax = get_glyph_cleaned_extents(
+                        glyf_table.glyphs[component.glyphName],
+                        glyf_table)
+
+                    if hasattr(component, 'transform'):
+                        transform = component.transform
+                        if not (transform[1][0] == transform[0][1] == 0):
+                          # handle 90-degree rotations and reflections
```
jamesgk: Did you mean to leave this in? If so should this be a todo?

nototools/render.py
```diff
@@ -113,12 +132,13 @@ def get_glyph_cleaned_extents(glyph, glyf_table):
 def get_glyph_vertical_extents(glyph_id, font_file_name):
     """Returns visible vertical extents given a glyph ID and font name."""
     font = font_caching.open_font(font_file_name)
-    glyf_table = font['glyf']
+    # glyf_table = font['glyf']
```
jamesgk: Commented line should be removed I think.

nototools/render.py
dougfelt: Oops, nope.

Issue #337, opened 2016-12-08
noto_lint.py:  Please modify noto_lint.py to handle omitted "Regular" in full name


marekjez86:
noto_lint.py expects for the full name 'Regular' to be present when there are regular weights, but the noto fonts all omit it. Please modify noto_lint.py to deal with omitted "Regular"

dougfelt:
Yeah, so, in phase 2 MTI omitted 'Regular' in both the full name and the postscript name, and Adobe included 'Regular' in both the full name and the postscript name.

Now the files omit 'Regular from the full name but include it in the postscript name.

@jamesgk, is this something in the sources from Monotype or something our toolchain does?

dougfelt:
FYI, the Glyphs-generated fonts include 'Regular' in both the full name and the postscript name.  Which is what lint had been expecting.  So it looks like the toolchain generates names differently than Glyphs.

jamesgk:
We haven't been including Regular in full names since this change to ufo2ft about a year ago:
https://github.com/googlei18n/ufo2ft/commit/0a7992919fc69661ef3c8941728e5642d9dfac34.

I don't really have a preference but I do generally trust Cosimo et al. to encode font designers' expectations into our tools.

dougfelt:
Yeah.  The thing is, 1) Adobe and MTI do it differently, and 2) our Glyphs-generated files differ in the naming from the tool-generated files.

The last I remember hearing from Behdad, he thought the glyphs sources should be changed where possible in preference to 'fixing' things in the toolchain.  This of course can be cumbersome in the case of a project like noto, where you want the ability to ensure consistent behavior across dozens of source files / hundreds of font instances.  Perhaps it is more straightforward now to apply automated edits across the glyphs sources?

jamesgk:
Actually, I don't think there is a way to force fontmake (or ufo2ft specifically) to include Regular in full names, no matter what we do to the source. If we really want this we might have to either change the default behavior of ufo2ft or make it configurable somehow.

Issue #336, opened 2016-12-07
noto_lint.py:  please teach noto_lint about NotoSansArabic.


marekjez86:
noto_lint doesn't know about NotoSansArabic and how its names should work.  please teach noto_lint about it.

marekjez86:
same applies to NotoSansArabicUI and NotoSansHebrew (I haven't checked yet about NotoSerifHebrew)

marekjez86:
same for NotoSansKhmer (I suspect also for NotoSerifKhmer)

dougfelt:
@marekjez86 I don't understand what your issue is w.r.t. Hebrew and Khmer.  Unlike with Arabic, noto_lint does have name data for these.  What exactly is the problem you're having with Hebrew and Khmer?

marekjez86:
For example for NotoSansKhmer I get:

Automatic testing for 'NotoSansKhmer-Thin.ttf', Version 1.901:
E <name/unable_to_check> No name data available for this font.

How do I resolve the above problem?
================================
Automatic testing for 'NotoSansHebrew-Thin.ttf', Version 1.902:
E <name/unable_to_check> No name data available for this font.

How do I resolve the above problem?

I though we need to "teach" noto_lint about NotoSansHebrew, NotoSansKhmer ... to resolve it but if there's some other way I'll be glad to work on fixing these errors.

dougfelt:
That's odd.  My repo has a change to these but I don't see how it got there, and it's not in the public repo.  My mistake.

The expected names was built originally from the phase 3 planning spreadsheet scope tab.  There we refer to these fonts as 'Arabic Sans New', 'Hebrew Sans New' and a bunch of others with '(new?)' because we never decided what they should be named, so just punted.

Now of course fonts are being delivered with whatever names someone most recently decided to use and lint has no idea.  This means, basically, that as new scripts are delivered lint will trip over this and need to be updated to deal with each late-breaking change.

Can you go over that list and determine what the names should be for the fonts we never quite made a decision on?

marekjez86:
You are right. That (* New) was the decision re new names. Thank you for the reminder. I'll go over the list. Taking bug ownership for now.

Issue #334, opened 2016-12-07
noto_lint.py: cannot parse anything with "ExtraLight" in a file name [a conjecture based on the sample data below]


marekjez86:
The pipeline produces files with such a name. We should be able to lint them also.

dougfelt:
So, currently we expect the name 'Thin' to be given to fonts with a weight of 250, because weights under 200 confuse some versions of windows.  And the old noto never had ExtraLight.

'ExtraLight' would normally be 200, and Thin would normally be 100.  ExtraLight should at least have a higher weight than Thin.

@jungshik, what should these names imply in terms of the  weight values?

dougfelt:
@marekjez86 please provide a list of all the various weight names we expect to be dealing with.

marekjez86:
Pipeline generates the following names:

Noto*-Black.ttf
Noto*-BlackItalic.ttf
Noto*-Bold.ttf
Noto*-BoldItalic.ttf
Noto*-Condensed.ttf
Noto*-CondensedBlack.ttf
Noto*-CondensedBlackItalic.ttf
Noto*-CondensedBold.ttf
Noto*-CondensedBoldItalic.ttf
Noto*-CondensedExtraBold.ttf
Noto*-CondensedExtraBoldItalic.ttf
Noto*-CondensedExtraLight.ttf
Noto*-CondensedExtraLightItalic.ttf
Noto*-CondensedItalic.ttf
Noto*-CondensedLight.ttf
Noto*-CondensedLightItalic.ttf
Noto*-CondensedMedium.ttf
Noto*-CondensedMediumItalic.ttf
Noto*-CondensedSemiBold.ttf
Noto*-CondensedSemiBoldItalic.ttf
Noto*-CondensedThin.ttf
Noto*-CondensedThinItalic.ttf
Noto*-ExtraBold.ttf
Noto*-ExtraBoldItalic.ttf
Noto*-ExtraCondensed.ttf
Noto*-ExtraCondensedBlack.ttf
Noto*-ExtraCondensedBlackItalic.ttf
Noto*-ExtraCondensedBold.ttf
Noto*-ExtraCondensedBoldItalic.ttf
Noto*-ExtraCondensedExtraBold.ttf
Noto*-ExtraCondensedExtraBoldItalic.ttf
Noto*-ExtraCondensedExtraLight.ttf
Noto*-ExtraCondensedExtraLightItalic.ttf
Noto*-ExtraCondensedItalic.ttf
[...]

Issue #333, opened 2016-12-07
noto_lint.py: allow also individual names in name/designer


marekjez86:
please allow the following string to pass:
E <name/designer> The designer is 'Danh Hong and the Monotype Design Team' but was expecting 'Monotype Design Team'
E <name/designer> The designer is 'Jelle Bosma - Monotype Design Team' but was expecting 'Monotype Design Team'

i.e., match anything with .\*Monotype Design Team.\* alternatively only \^Monotype Design Team.\* OR .\*Monotype Design Team$

PR #345, opened 2016-12-14
Lint regular fix

### LINKED ISSUE #344: Lint should ignore off-curve extrema errors under 1 unit.
An example current error:

E <paths/extrema> The glyph 'uni17D21791' is missing on-curve extreme points in the segment between point 24=(-435, -161) and point 26=(-409, -174) by 0.473684 units.

According to Werner Lemberg this won't affect autohinting at normal point sizes:

"As far as I can see, the deviations from the extrema are <= 1 font
unit.  This is completely irrelevant at the normal hinting sizes
(i.e., <= 40ppem approx.) – in other words, TTFautohint won't have any
problems with those glyphs, I believe."

dougfelt:
A series of individually-small changes to make noto_lint more compatible with phase_3 fonts.  Also includes a few unrelated changes (notoconfig, quiet warnings). See individual commits for more info.

Should fix #344, #337, #336, #334, #333.

### DIFF
--- a/nototools/noto_lint.py
+++ b/nototools/noto_lint.py
@@ -197,8 +197,11 @@ def out_of_box_size(curve):
     # Bounding box of the bezier will contain that of the endpoints.
     # The out-of-box size for the entire curve will be maximum of the deviation
     # for the first curve and that of the remaining curve.
-    return max(ex1 - bx1, ey1 - by1, bx2 - ex2, by2 - ey2,
+    delta = max(ex1 - bx1, ey1 - by1, bx2 - ex2, by2 - ey2,
                out_of_box_size(remaining_curve))
+    # ignore very small deviations
+    return 0 if delta < 1 else delta
+
 
 def calc_bounds(piece):
     if len(piece) == 2:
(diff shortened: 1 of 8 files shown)


nototools/noto_lint.py
```diff
@@ -771,7 +769,15 @@ def check_cmap_table():
                      "which it shouldn't have." % (
                          table.format, table.platformID, table.platEncID))
             elif table != (12, 0, 4):
-                cmaps[table.format] = table.cmap
+                if table.format in cmaps:
+                    # if we have both 4,3,1 and 4,0,3, they should be aliases
+                    # similarly if we have both 12,3,10 and 12,0,4
+                    if id(table.cmap) != id(cmaps[table.format]):
+                        warn("cmap/tables/notaliased", "cmap",
```
jamesgk: Should this be `cmap/tables/unaliased`?

dougfelt: Um, whoops.  All the fonts I ran this on pass so I didn't catch that.

PR #347, opened 2016-12-15
Catch subsetter failure in lint and report.

dougfelt:


nototools/noto_lint.py
```diff
@@ -1823,7 +1823,12 @@ def check_accessiblity(cmap):
         all_glyphs = set(font.getGlyphOrder())
         subsetter = subset.Subsetter()
         subsetter.populate(unicodes=cmap.keys())
-        subsetter._closure_glyphs(font)
+        try:
+          subsetter._closure_glyphs(font)
+        except Exception as e:
```
jamesgk: Could we catch a more specific exception?

dougfelt: Why?

jamesgk: Well, I'm generally wary about catch-all exception blocks that don't later raise the exception. I figure it's less likely we'll be surprised by something down the road if we only catch an exception we expect to be raised. But I suppose since we always alert that something was caught this isn't so bad.

dougfelt:
Yeah, well, we don't expect this to raise any exception at all.  There's
some bad table in the font, is my guess-- this is a 'list index out of
range' IndexError but I expect other kinds of table errors could occur as
well.  I just want lint to be able to continue for now.  I do need to
investigate, perhaps the subsetter is at fault and this code path hasn't
been exercised before.

PR #350, opened 2016-12-20
hb_input speedup

jamesgk:


nototools/hb_input.py
```diff
@@ -191,11 +191,12 @@ def _input_from_5_1(self, gsub, st, glyphs, target_i, cur_i, fea, seen):
         inputs = []
         for ruleset in st.SubRuleSet:
             for rule in ruleset.SubRule:
+                if (not any(subst_lookup.LookupListIndex == target_i
```
dougfelt: Do you need the outer parens?

nototools/hb_input.py
```diff
@@ -99,13 +103,15 @@ def input_from_name(self, name, features=(), seen=None, pad=False):
         # but we should avoid returning None here if there are other options
         inputs = [i for i in inputs if i is not None]
         if not inputs:
-            return None
+            self.memo[name] = None
+            return self.memo[name]
```
dougfelt: maybe use 'else' for the remainder, and then just the one return line?

PR #356, opened 2017-01-11
[notodiff] Fairer qualification of render diff

jamesgk:
With this change, if two pixels between renderings differ by less than
255 (i.e. one or both is a shade of gray), the pixel difference counts
less towards the overall difference. This should reduce the amount of
noise from glyphs which have changed just a slight amount and whose
rendering difference is barely visible.


marekjez86:
LGTM

PR #363, opened 2017-01-27
Make notoconfig.py silent.

dougfelt:
Tools that depend on nototools files and that emit output to stdout
for other code to consume don't want stdout polluted with these
messages.


anthrotype:
or you could `print("message", file=sys.stderr)` or `sys.stderr.write("message\n")`.

PR #364, opened 2017-01-31
Update cmap data.

dougfelt:
This makes three main changes:
- Monotype reported that a number of Latin punctuation chars in Lao didn't
  need special handling and could be handled by fallback to LGC.  This
  removes those characters.
- A comment on noto-fonts#669 noted that U+2195 was used by some
  git tools, and so it should be supported in MONO.  Other characters it
  mentions were already planned for MONO.
- Android doesn't want to fallback to MONO just to get full coverage, so
  didn't like the previous plan to support APL only in the MONO font.  This
  once again duplicates the APL set in Math and Mono.

Also, a transient update to the emoji-data file made U+263F an emoji, so it
is also added to Emoji here.  However this data file has changed in the last
two weeks and the character is no longer emoji.  Rather than update the
emoji data files just for this change, we'll wait until they stabilize a bit,
and leave the cmap data as it is for now.

This is the result of compare_cmap_data:
base: 2016-12-19 noto_cmap_reqs mergefile=data/noto_cmap_phase3.xml
target: 2017-01-31 noto_cmap_reqs mergefile=mti_cmaps_170131.xml
Laoo # Lao
  removed, now in fallback (18): 0021-0022 0027-0029 002c-002f 003a-003b 003f 00ad 2018-2019 201c-201d 2026
MONO # MONO
  added (1): 2195
Zmth # Math
  added (69): 2336-236d 236f-237a 2395
Zsye # Emoji
  added (1): 263f


dougfelt:
@jungshik, tried following markdown to reference the issue (in noto-fonts) but it's not highlighting for me.

PR #370, opened 2017-02-11
Fix a typo in render.py

jungshik:
glyph_set => glyf_set to make render.py work


nototools/render.py
```diff
@@ -123,7 +123,7 @@ def get_glyph_cleaned_extents_OLD(glyph, glyf_table):
 def get_glyph_vertical_extents(glyph_id, font_file_name):
     """Returns visible vertical extents given a glyph ID and font name."""
     font = font_caching.open_font(font_file_name)
-    glyph_set = font.getGlyphSet()
```
behdad: I think `glyf_set` should have been renamed to `glyph_set` instead...

PR #375, opened 2017-02-28
Add data on hinted scripts and ttfautohint script codes.

dougfelt:


dougfelt:
ttfautohint hints two kinds of Georgian, Mkhedruli and Khutsuri.  Which is ours?

PR #378, opened 2017-03-02
Add autofix for phase 3 tool.

dougfelt:
This is a quick oneoff tool to autofix and hint binaries for phase 3.
The idea is to take selected phase 3 fonts built by the pipeline
and promote them to the main noto-fonts hinted/unhinted directories.

Since we have existing hinted versions of some fonts, we generate
autohinted versions where necessary so both hinted and unhinted are from
the same source.

We write information into the version string so we can identify where
the font binary came from.  Currently, the binaries are built
from source and then checked in to the alpha branch of noto-fonts
without any information about the version of noto-source or the build
tools.  So the best we can do is to reflect the state of noto-fonts at
the time we took the fonts for processing.

### DIFF
--- a/nototools/autofix_for_phase3.py
+++ b/nototools/autofix_for_phase3.py
@@ -1,0 +1,270 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


nototools/autofix_for_phase3.py
```diff
@@ -0,0 +1,270 @@
...
+  return font_info
+
+def _is_ui_metrics(f):
+  return _get_font_info(f).is_UI_metrics
+
+
+def _autohint_code(f, script):
+  """Return 'not-hinted' if we don't hint this, else return the ttfautohint
+  code, which might be None if ttfautohint doesn't support the script.
+  Note that LGC and MONO return None."""
```
jungshik: So, we don't autohint LGC and Mono? Or, do you mean 'no-script'?

dougfelt: Oops, thanks.  I came up with no-script after writing this but I forgot to back and change it.

PR #381, opened 2017-03-08
Add merge_fonts.py and substitute_linemetric.py to nototools

zjusbo:
These two scripts are used to merge noto fonts and substitute line metrics using those in LGC font.

merge_fonts partially depends on merge_noto but they differs in:
1. the font source of merge_fonts are maintained in a pure list in script. There is no filename manipulations, splitting, construction stuff in it.
2. The line metrics of the result font is substituted by a specific font. This is to achieve a visual harmony for different scripts in a font.


zjusbo:
As talked with @dougfelt , It's time to move test cases to separate folder under nototools/tests. I'm opening another PR to move it.

PR #382, opened 2017-03-08
Move test cases to a separate folder

zjusbo:


dougfelt:
Be careful.  The original data directory belongs under nototools and should retain its original contents.  Just don't put your test data there.  Your test data files, and only those files, belong in tests/data.  Other data stays where it is.

zjusbo:
Done. I thought it was test data. PTAL.

PR #383, opened 2017-03-08
Create Bopo and Tglg sample texts

Kanjilearner817:
I didn't see sample texts for either script, so I thought that I would them myself.


Kanjilearner817:
I signed it!

PR #384, opened 2017-03-09
Small tweak to exemplar sample generation code, add some char samples.

dougfelt:
- Char sample names now include '_char'.

These samples aren't used by the website code but it seems reasonable to
have them.


dougfelt:
I probably want to update the CLDR data (it's two years old) and rerun the tool.  Earlier this morning I accepted a BoPoMoFo sample, the repo currently doesn't have exemplar data for one.  Changing the CLDR data could impact a lot of things so I want to wait a few weeks before I try that.

PR #385, opened 2017-03-09
Generate and include README files when building packages.

dougfelt:
This will fix noto-fonts#854

- Some tool fns were added to issue git command lines from python
and extract the required info.

- README files are generated using this info, and the corresponding
ones added to the zips when they are built.

- In the process of collecting info for the README files, we also
check to see if the repos are at a tagged commit (since we include
the tag text), and by default raise an exception if the repos are
not (or have pending changes).

Other minor changes:
- reorder the zip building processes when possible to put the
README and LICENCE files first.
- add a flag to allow continuing when repos don't pass the check
- change the name of a flag

### DIFF
--- a/nototools/generate_website_2_data.py
+++ b/nototools/generate_website_2_data.py
@@ -49,13 +49,18 @@ from nototools import unicode_data
 TOOLS_DIR = notoconfig.noto_tools()
 FONTS_DIR = notoconfig.noto_fonts()
 CJK_DIR = notoconfig.noto_cjk()
+EMOJI_DIR = notoconfig.noto_emoji()
 
 CLDR_DIR = path.join(TOOLS_DIR, 'third_party', 'cldr')
 
 LAT_LONG_DIR = path.join(TOOLS_DIR, 'third_party', 'dspl')
 SAMPLE_TEXT_DIR = path.join(TOOLS_DIR, 'sample_texts')
 
-APACHE_LICENSE_LOC = path.join(FONTS_DIR, 'LICENSE')
+# The Apache license is currently not used for our fonts, but leave
+# this just in case this changes in the future.  We have a copy of
+# the Apache license in the noto-emoji repo, since it covers the
+# code there, so just use that.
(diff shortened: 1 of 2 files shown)


nototools/generate_website_2_data.py
```diff
@@ -59,6 +58,13 @@
 APACHE_LICENSE_LOC = path.join(FONTS_DIR, 'LICENSE')
```
jungshik: Is FONTS_DIR the top of noto-fonts dir? If so,  LICENSE file there is also OFL, isn't it? 
LICENSE file in noto-emoji is Apache, but for all other fonts, LICENSE is OFL 1.1.

nototools/generate_website_2_data.py
```diff
@@ -678,13 +725,16 @@ def build_zips(self, families):
   def build_universal_zips(self, families):
     hinted_fonts = []
     unhinted_fonts = []
+    readme_path = self.get_readme_path('all')
     for family_data in families.values():
       hinted_fonts.extend(
           family_data.hinted_members or family_data.unhinted_members)
```
jungshik: Does this mean that "Noto-hinted zip" will have both hinted and unhinted fonts?

dougfelt: Hmmm.  It has for a few years now, I think.  It's not so much "these are the hinted fonts" as it is "these are fonts we suggest if your platform does better with hinted fonts".  So it provides hinted fonts where we have them, and unhinted if not, so that it provides the same coverage.  This does mean the file name is misleading.

I think originally we had three sets of fonts and one was named for windows, which is arguably cleaner naming.

nototools/generate_website_2_data.py
dougfelt: Akk.  Yeah, this predates the switch to use OFL.

Currently, all the fonts use the OFL license, but if any did use Apache they'd get the wrong one, so I still need to fix this.  I'll use the one at the top of noto-emoji as you suggest.

noto-emoji is weird, the code is Apache but the fonts are OFL.  We have a different license in the fonts subdirectory than at the top level...

PR #386, opened 2017-03-10
Android patch update

dougfelt:
Changes to the android patch tool and some utilities it uses.

### DIFF
--- a/nototools/android_patches.py
+++ b/nototools/android_patches.py
@@ -16,24 +16,26 @@
 
 """Patches for Android versions of Noto fonts."""
 
+import argparse
 import codecs
 import glob
+import os
 from os import path
+import shutil
+import tempfile
 
 from nototools import subset
 from nototools import coverage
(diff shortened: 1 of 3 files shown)


nototools/android_patches.py
```diff
@@ -114,17 +121,47 @@ def _remove_from_cmap(infile, outfile, exclude=[]):
...
     print 'remove cjk emoji', font_name
     _remove_from_cmap(
-        font_name,
-        path.join(outdir, font_name),
+        path.join(srcdir, font_name),
+        path.join(dstdir, font_name),
         exclude=EMOJI)
 
 
+def patch_cjk_ttc(ttc_srcfile, ttc_dstfile):
```
jungshik: CrOS may need different changes than Android. Not now, but eventually we may have to move this and other 'generic' functions (not tied to Android) out of android-patches.py.

dougfelt: Sure, just let me know what the different requirements are.

PR #388, opened 2017-03-17
Compressed ranges

dougfelt:
Add a new syntax for the utility method that parses ranges from a text representation, and use it.

Also fixed a small issue in the tool used to generate character comparison review docs.

### DIFF
--- a/nototools/tool_utils.py
+++ b/nototools/tool_utils.py
@@ -314,13 +314,83 @@ def svn_update(repo):
 
 
 def parse_int_ranges(
-    range_string, is_hex=True, sep=None, allow_duplicates=False):
-  """Returns a set of ints from a string of numbers or ranges separated by sep.
-  A range is two values separated by hyphen with no intervening separator;
-  ranges are inclusive."""
-  result = set()
-  count = 0
+    range_string, is_hex=True, sep=None, allow_duplicates=False,
+    return_set=True, allow_compressed=False):
+  """Returns a set/list of ints from a string of numbers or ranges separated by
+  sep.  A range is two values separated by hyphen with no intervening separator;
+  ranges are inclusive.  If allow_compressed is true, '/' is also allowed
+  as a separator, and ranges following it or hyphen are interpreted as suffixes
+  that replace the same number of characters at the end of the previous value.
+  '-' generates the range of intervening characters as before, while '/' does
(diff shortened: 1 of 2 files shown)


nototools/tool_utils.py
```diff
@@ -314,45 +314,118 @@ def svn_update(repo):
...
+  ranges are inclusive.  If allow_compressed is true, '/' is also allowed
+  as a separator, and ranges following it or hyphen are interpreted as suffixes
+  that replace the same number of characters at the end of the previous value.
+  '-' generates the range of intervening characters as before, while '/' does
+  not.  Returns a set or a list depending on return_set."""
+
   base = 16 if is_hex else 10
+  vals = []
+
+  def _add_segment(prev_str, suffix, is_range):
```
jungshik: Just a thought (no need to change):  Isn't  parse_int_ranges() too long? Wouldn't it be better to move two nested functions out of it?

dougfelt: I'd rather not.  They're specific to this function, and I like the ability to scope access to functions like this.  They also have access to variables in their enclosing scope.  Moving them outside means they require more parameters and their (sole) client isn't as clear without further documentation.  Until there's another client I'd rather keep them here.

nototools/tool_utils.py
```diff
@@ -314,45 +314,118 @@ def svn_update(repo):
...
-  count = 0
+    range_string, is_hex=True, sep=None, allow_duplicates=False,
+    return_set=True, allow_compressed=False):
+  """Returns a set/list of ints from a string of numbers or ranges separated by
+  sep.  A range is two values separated by hyphen with no intervening separator;
+  ranges are inclusive.  If allow_compressed is true, '/' is also allowed
+  as a separator, and ranges following it or hyphen are interpreted as suffixes
+  that replace the same number of characters at the end of the previous value.
+  '-' generates the range of intervening characters as before, while '/' does
+  not.  Returns a set or a list depending on return_set."""
```
dougfelt: Added an example.  Yes, it's difficult to understand without some examples.  I hope one will suffice, there are variations and edge cases but they'd require more examples than I think is warranted.  Let me know though if you'd like some more.

nototools/tool_utils.py
jungshik: Thanks a lot. That was sufficient.

PR #394, opened 2017-03-29
Math cmap update2

dougfelt:
Update to tool and tool-generated data for Math cmap.

Also tweak the tool that generates comparison images to
- display lists of characters in LTR order regardless of native order.
- report hex values of codepoints that fail due to missing unicode name data


nototools/generate_dingbats_html.py
```diff
@@ -864,7 +864,10 @@ def _flagged_name(cp, flag_sets):
   is a map from flag name to a tuple of cp set and boolean.
   True means add flag if cp in set, False means add flag if it is
   not in the set."""
-  name = unicode_data.name(cp)
+  try:
+    name = unicode_data.name(cp)
+  except:
+    raise Exception('no name for %04X' % cp)
```
jungshik: just curious, what cp does not have name in unicode_data?

dougfelt: Typos don't.  For example if I specify a unicode range and it includes undefined codepoints by mistake.

PR #395, opened 2017-03-31
Enhancement tweaks

miguelsousa:


miguelsousa:
@dougfelt @jungshik any interest in these changes?

dougfelt:
Sorry, been focussing on other things at the moment.  Meant to respond to this.

I seem to recall seeing a message go by on the fontmake repo saying they have fixed version numbers because allowing them to vary can introduce issues.  If I understand right this change allows any version newer than the one specified to be used.  In nototools this might not be as much of a problem, though.

miguelsousa:
@dougfelt I ran into bugs in the pinned versions of booleanOperations and Pillow, which went away once I allowed them to be updated. That's what got me to do the same for the other dependencies. The referenced versions are quite old for most of them anyway. I can pin them to the versions I have installed, if you prefer. Below is what I'm currently running. List made with `pip freeze`.
```
booleanOperations==0.7.0
defcon==0.3.1
fonttools==3.9.1
Pillow==4.0.0
pyclipper==1.0.6
ufoLib==2.0.0
```

anthrotype:
Or you could have a bot like https://pyup.io/ periodically update requirements.txt for you, which is what fontmake does.

miguelsousa:
@anthrotype totally agree, but that's a decision for @dougfelt to make.

dougfelt:
@miguelsousa thanks, why don't you pin them to your current versions for now.  An auto-update script would help but we'd need better automated unit testing for that, I think.

miguelsousa:
@dougfelt done

PR #396, opened 2017-03-31
Add patch_post_table to android_patches: post table 2.0 => 3.0

jungshik:
Android wants to save space by changing post version 2.0 to 3.0.
A few kBs per font can be saved this way.


dougfelt:
Hmmm... each of the patch fns copies the updated font to dstdir.  The existing patch fns get away with this because they each operate on different fonts.  But the new patch for the post table could overwrite any files that had hyphens added to them by that earlier patch operation.  The architecture for the utility is a poor fit for cases where a single font might have multiple patches applied...

jungshik:
Ooops...

dougfelt:
Were you going to change this?

PR #400, opened 2017-04-04
Website update

dougfelt:


nototools/generate_website_2_data.py
```diff
@@ -554,7 +554,7 @@ def get_css_generic_family(family):
 
 CSS_WEIGHT_MAPPING = {
     'Thin': 250,
```
jungshik: What does this mapping do? Shouldn't Thin be mapped to 100 (even though inside the font, OS/2.fsWeight is set to 250) for CSS?

nototools/generate_website_2_data.py
```diff
@@ -994,10 +996,17 @@ def build_family_images(
...
-        family_suffix = ' ' + font.weight
+      if is_cjk and family.name.find('Serif') < 0:
+        # The sans and serif cjk's are named differently, and it confuses
+        # fontconfig.  Sans includes 'Regular' and 'Bold' in the standard
+        # font names, but serif doesn't.  Fontconfig registers two names
+        # for these in the sans (with and without the weight), but only
+        # the name without weight for the serif.  So if you ask pango/cairo
+        # for the serif font and include the weight name, it fails and
+        # falls back to some non-noto font.
+        family_name = family.name + ' ' + font.weight
```
jungshik: Yeah, this is what Ken did to style-lock Serif (he just did it for Regular and Bold).

nototools/generate_website_2_data.py
dougfelt: This maps the name to a weight value in the font.

nototools/generate_website_2_data.py
dougfelt: Yes, this is what one would normally expect.  The sans are the odd ones.  Unfortunately, I had applied the workaround for the sans to all the CJK.

jungshik: In an ideal situation, all the weights should be style-locked, but the 'history' of name ids makes it harder as you know. Ken wants to move intentionally slowly on this.

nototools/generate_website_2_data.py
jungshik: In that case, ```'ExtraLight': 200``` is odd. Does ExtraLight have OS/2.fsWeight set to 200? 

I thought we use '250' in OS/2.fsWeight [1] because Windows GDI does not work well with anything lower than 250.  That being the case, ExtraLight with OS/2.fsWeight of 200 wouldn't work, either. No? 

[1] I'm not a fan of using 250 for Thin to make Windows GDI happy because Android, Chrome OS, Mac and Linux work perfectly well with 100.

dougfelt: Sigh.  I should have looked in the font.  Xiangye said it should be 'ExtraLight 200' so that's what we did.  Only the name should have changed, the weight should have remained '250' according to this.

nototools/generate_website_2_data.py
dougfelt: I thought you couldn't style-lock all the weights via the family name, only regular/bold, and italic/bolditalic.  The other name fields are populated (16/17) so I'm not sure what else can be done.

nototools/generate_website_2_data.py
jungshik: This PR is about web-site update. Sorry if I'm slow. 
Do you use these values in CSS?  In that case, 100 for Thin and 200 for ExtraLight should just work fine and those values are what to use in CSS. Only inside a font file, we don't use anything lower than 250 for OS/2.fsWeight to make GDI happy.

dougfelt: No, we use them to label the font weight with additional information, and to coordinate the samples with the font weight.

Noto Sans CJK DemiLight internally is 350 and is intermediate between Light and Regular, which are 300 and 400 respectively.  There's no existing CSS value for DemiLight, nor is there a name for it.

jungshik: Ok. 
I know of DemiLight (350). It's added to match the lighter of Roboto and Noto Sans (I forgot which) while 400 match the thicker of the two. 
Anyway, did you change ExtraLight to 250 as well?  With that, LGTM

nototools/generate_website_2_data.py
jungshik: By "style-locking" all the weights, I meant setting 16 and 17 properly so that all the weights are treated as belonging to a single family. (the limitation of ids=1/2 I'm aware of).
I found it that they're in Noto Serif CJK.  The difficulty I have in mind was that not every OS/library refers to id=16/17. Perhaps, it's better now than in the past.

dougfelt:
Well, we do generate CSS with these values, but it seems we can't be using these files because the Sans uses 250 for Thin and 350 for DemiLight.

dougfelt:
I'm thinking I should change these to the 'ideal values', e.g. how we'd assign them if it weren't for the GDI issue forcing the minimum os/2 weight to be 250.  'Thin' would get weight 100, 'ExtraLight' would get weight 200.  'DemiLight' would still use '350' (for display to users, and identifying the sample file) since it is, in fact, intermediate in weight between Light and Regular, but the CSS @font-face will have to do something different for it.  

My thought is to generate @font-face using weight 350 (as we do now) for environments that are loose and/or looking forward to the variable-font world, and also generate another @font-face with the same url but name it with the weight explicitly, e.g. 'Noto Sans CJK TC 350'.  So a web page could still get this font using this face name (of course using 'lighter'/'darker' or setting weight on this face would not work as expected).  This would allow an html page displaying sample text using web fonts to access this font by coding the name differently, as a hack to work around the current CSS limitations on weight.

nototools/generate_website_2_data.py
```diff
@@ -754,30 +740,37 @@ def copy_font(self, fontpath):
...
+          css_family += ' ' + str(weight)
+          # prevent auto-bolding of this font by describing it as bold
+          weight = 700
+        slope = css_style(font.slope)
         css_file.write(
           '@font-face {\n'
           '  font-family: "%s";\n'
           '  font-weight: %d;\n'
           '  font-style: %s;\n'
           '  src: url(../fonts/%s) format("truetype");\n'
```
jungshik: I don't what browsers would do with format set to 'truetype' for CFF fonts. 

BTW, where is this CSS used? Internal only or for get/noto page?  If it's the latter, shouldn't we have woff2 and woff compressed versions instead of otf/ttf?

dougfelt: The spec says that 'opentype' and 'truetype' are to be considered synonymous and that neither has implications either for the format of the outlines or the presence of opentype tables.

dougfelt:
I'm not sure whether the css is used at the moment, I'll ask.

nototools/generate_website_2_data.py
dougfelt: https://www.w3.org/TR/css-fonts-3/#src-desc

jungshik: Ok. Thank you !

If CSS is not actually used yet,   LGTM (perhaps with TODO for woff/woff2).

dougfelt:
It is not.

jungshik:
> It is not.

Then, please go ahead and merge. (Somehow I don't see 'Review change' button so that I can't give a formal approval).

PR #406, opened 2017-04-13
Support legacy Android sequences.

dougfelt:
Since Android shipped with support for these sequences, they want to
keep them around even though they're no longer in the unicode emoji
data.

This hard-codes information about the sequences into unicode_data
in the form of text, and changes some code to use them by default.

_LEGACY_ANDROID_SEQUENCES takes the form of the emoji sequence data
and is parsed by the same code, except reworked slightly so that
the text to be parsed can be provided as a list of lines as well
as a filename.

_LEGACY_ANDROID_ORDER is a list of sequences.  Each sequence starting
with '-' is a key, the following are sequences to be added after
the key.  This is used to insert these sequences into the order
data returned by parsing the emoji-test data file.  For example,
'men wrestling' sequences with skin color are inserted after the
'men wrestling' sequence.  This keeps the patch out of the
unicode-supplied data file.

The order data is patched using two functions, one to define a map
of patch data, one to apply it, because who knows but we'll have to
patch more stuff in the future and this seemed cleaner, somehow.

Ideally of course we'd provide a way to get the unpatched data, but
for now the android requirements are driving this.

### DIFF
--- a/nototools/unicode_data.py
+++ b/nototools/unicode_data.py
@@ -745,28 +745,32 @@ EMOJI_SEQUENCE_TYPES = frozenset([
     'Emoji_ZWJ_Sequence',
     'Emoji_Single_Sequence'])
 
-def _read_emoji_data_file(filename):
-  """Parse an emoji data file and return a map from sequence to tuples of
+def _read_emoji_data(lines):
+  """Parse lines of emoji data and return a map from sequence to tuples of
   name, age, type."""
   line_re = re.compile(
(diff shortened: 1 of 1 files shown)


nototools/unicode_data.py
```diff
@@ -885,6 +1017,11 @@ def _load_emoji_group_data():
   with open_unicode_data_file('emoji-test.txt') as f:
     text = f.read()
   group_list = _read_emoji_test_data(text)
+
+  # patch with android items
+  patch = _get_android_order_patch()
+  group_list = _apply_order_patch(patch, group_list)
```
jungshik: just a thought: 'order' might as well be dropped from both function names above.

PR #410, opened 2017-05-04
Extend lookup path for notoconfig.

dougfelt:
Now also looks in /usr/local/share/noto if it exists, so a
non-user-specific location is available for the notoconfig data.

### DIFF
--- a/nototools/notoconfig.py
+++ b/nototools/notoconfig.py
@@ -40,20 +40,20 @@ def _setup():
   values will hold a mapping from the <name> to value.
   Blank lines and lines starting with '#' are ignored."""
 
-  configfile = path.expanduser("~/.notoconfig")
-  if path.exists(configfile):
-    with open(configfile, "r") as f:
-      for line in f:
-        line = line.strip()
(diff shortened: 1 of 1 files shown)


nototools/notoconfig.py
```diff
@@ -40,20 +40,21 @@ def _setup():
...
-        k, v = line.split('=', 1)
-        values[k.strip()] = v.strip()
-  else:
-    # This needs to be silent.  It causes a makefile error in noto-emoji,
-    # which expects stdout to consist only of the output of a python
-    # script it runs.
-    pass
+  roots = ['~', '/usr/local/share/noto']
+  for root in roots:
+    configfile = path.expanduser(path.join(root, '.notoconfig'))
```
jungshik: A dot file in /usr/local/share/noto is a bit odd, isn't it?  How about /usr/local/share/noto/notoconfig or /usr/local/share/noto/config ?

dougfelt: Yeah, it is.  Ok I'll change that.

PR #420, opened 2017-05-19
Docker tooling

dougfelt:
First pass.  See commit msg for more info.

### DIFF
--- a/nototools/docker/make_getnoto/build_noto_site_data.sh
+++ b/nototools/docker/make_getnoto/build_noto_site_data.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# Build data for get/noto site.
+
+# Map /tmp/website2 to an external location to get the data when this is
+# complete.
+
+# Put the repos at the latest tagged release.  This does not do a git pull, we
+# assume noto is already at or beyond the latest release.  Might want to rethink
+# that.
+cd /app/noto/nototools/nototools
+./sync_repos.py -v
+
+# Add missing font package to the CJK repo.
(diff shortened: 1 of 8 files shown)


nototools/docker/make_getnoto/build_noto_site_data.sh
```diff
@@ -0,0 +1,54 @@
...
+# Put the repos at the latest tagged release.  This does not do a git pull, we
+# assume noto is already at or beyond the latest release.  Might want to rethink
+# that.
+cd /app/noto/nototools/nototools
+./sync_repos.py -v
+
+# Add missing font package to the CJK repo.
+# This is too big for github.  Currently we fetch it from the website itself,
+# which is kind of circular, since we're building the website...  We need a 
+# 'prerelease' location for things like this.
```
jungshik: How about  Google Cloud Storage (BigStore) for this?  Chrome OS team uses it for storing Gentoo source tar balls and Chrome team uses it to store test results / logs  among other things.

dougfelt: Yeah, this comment is not so clear.  This is actually fetching the zip from gcs mapped through our website, and putting it into the cjk dir.  Then we process the cjk dir files, including this, and write to our local staging directory.  From there it will later get pushed to gcs (not in this script).

This is using public urls so I don't need to pass any authentication info through to the script (wasn't sure how to best set that up-- of course I should figure that out).

davelab6:
@graphicore @felipesanches you might like to take a look at this :)

felipesanches:
thanks, @davelab6 !

PR #421, opened 2017-05-23
Add Pillow to requirements.txt

behdad:
In an environment I didn't have PIL installed and "pip install pil' didn't help.  Took some googling to find out that Pillow is the currently maintained fork of (dead) PIL.


dougfelt:
Pillow is already in requirements.txt.  Was the version incorrect?

behdad:
Err. Nevermind.  I first looked for PIL in requirements, didn't see it, then later when found Pillow didn't notice that it was in requirements already. Oops.

PR #426, opened 2017-05-31
Handle case where no notoconfig file exists.

dougfelt:


dougfelt:
Created the change on my new machine before I had git completely set up...

PR #429, opened 2017-06-15
Add Bhaiksuki script to is_indic check

punchcutter:
Add Bhaiksuki to the is_indic check since it is most definitely Indic.

### DIFF
--- a/nototools/create_image.py
+++ b/nototools/create_image.py
@@ -94,7 +94,7 @@ class DrawParams:
                  language=None, rtl=False, vertical=False,
                  width=1370, font_size=32, line_spacing=50,
                  weight=pango.WEIGHT_NORMAL, style=pango.STYLE_NORMAL,
-                 stretch=pango.STRETCH_NORMAL):
+                 stretch=pango.STRETCH_NORMAL, maxheight=0, horiz_margin=0):
         self.family = family
         self.language = language
         self.rtl = rtl
@@ -105,6 +105,8 @@ class DrawParams:
         self.weight = weight
         self.style = style
         self.stretch = stretch
+        self.maxheight = maxheight
+        self.horiz_margin = horiz_margin
 
     def __repr__(self):
         return str(self.__dict__)
@@ -176,6 +178,10 @@ def draw_on_surface(surface, text, params):
     # Newly-encoded scripts can be newer than your libraries and will
     # likely order LTR if you implementation doesn't know about them.
 
+    # The width is the desired width of the image.  The layout uses this
+    # width minus the margin.
+    width = params.width - 2 * params.horiz_margin
+
     font = pango.FontDescription()
     font.set_family(params.family)
     font.set_size(params.font_size * pango.SCALE)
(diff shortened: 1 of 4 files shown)


dougfelt:
The issue is not whether it is Indic per se, but if the checks influenced by this flag in noto_lint should behave the same for Bhaiksuki as they do for the current scripts for which is_indic returns true.

Should they?

punchcutter:
Yes, I see not all Indic scripts are in the list. I really only noticed because for Bhaiksuki if I run noto_lint as is on OS X with a narrow build of Python it fails because of upper plane characters (everything in Bhaiksuki) being passed to `ord()` in places like `unicode_data.name()`. Adding Bhks to the list made noto_lint finish, but just to be sure I also compared that with a Linux wide build (without adding to `is_indic`) and the difference is
```
W <gdef/classdef/not_combining_mismatch> Glyph AA_vwl (U+11C2F BHAIKSUKI VOWEL SIGN AA) is defined as class 3 (non-spacing) in the GDEF/GlyphClassDef table, but is of general category Mc.
W <gdef/classdef/not_combining_mismatch> Glyph E_vwl (U+11C38 BHAIKSUKI VOWEL SIGN E) is defined as class 3 (non-spacing) in the GDEF/GlyphClassDef table, but is of general category Mc.
W <gdef/classdef/not_combining_mismatch> Glyph AI_vwl (U+11C39 BHAIKSUKI VOWEL SIGN AI) is defined as class 3 (non-spacing) in the GDEF/GlyphClassDef table, but is of general category Mc.
W <gdef/classdef/not_combining_mismatch> Glyph O_vwl (U+11C3A BHAIKSUKI VOWEL SIGN O) is defined as class 3 (non-spacing) in the GDEF/GlyphClassDef table, but is of general category Mc.
W <gdef/classdef/not_combining_mismatch> Glyph AU_vwl (U+11C3B BHAIKSUKI VOWEL SIGN AU) is defined as class 3 (non-spacing) in the GDEF/GlyphClassDef table, but is of general category Mc. 
```
I'm not sure why all of these were encoded as Mc because only one is really spacing. Being defined as class 3 is because they behave like nonspacing marks for the most part and in almost all situations form ligatures and conjuncts. 

Also related, 200b and 25cc are required to be in the font, but aren't in noto_cmap_phase3.xml yet.

dougfelt:
Thanks for the clarification.

The noto tools are not really safe for use on narrow python builds, we don't test or use them on narrow builds and the attempts to make them work in those environments have been sporadic and incomplete at best.  Sorry about that.

The warnings you see are generally informative and not conclusive.  Often glyphs are assigned a class at the convenience of the designer based on how they prefer to set up the shaping tables, the tests however assume the glyphs are assigned a class compatible with their unicode properties.  In the past we've been given reports by the designers when the reports are inappropriate and we disable them using the lint config file.  I'll ask about these characters.

Noto fonts are designed to be used together with LGC as a fallback and so typically (not always true in the past) don't include characters that can be provided via fallback to LGC.  I expect 200b is in this category for a few reasons-- it doesn't render and serves primarily to provide a line break opportunity and stop shaping/combining behavior, which will happen if used via fallback.  That said we don't want to make life unnecessarily difficult in environments that don't provide fallback, so perhaps this should be added if it is commonly used.

I unfortunately did not make a thorough pass over all the fonts to ensure that fonts with combining characters contained 25CC and displayed those characters  properly in conjunction with it.  It is our intent to do so, so this is clearly an error.

I'll add these two characters to the requested characters list for Bhaiksuki.  Thanks for reporting them.

punchcutter:
> The warnings you see are generally informative and not conclusive. Often glyphs are assigned a class at the convenience of the designer based on how they prefer to set up the shaping tables, the tests however assume the glyphs are assigned a class compatible with their unicode properties. In the past we've been given reports by the designers when the reports are inappropriate and we disable them using the lint config file. I'll ask about these characters.

Currently these are defined as class 3 to work with other marks in the same mark attachment and mark filter classes since they act like other above base marks. I'm not bothered by the warnings, but they could be suppressed. 

> Noto fonts are designed to be used together with LGC as a fallback and so typically (not always true in the past) don't include characters that can be provided via fallback to LGC. I expect 200b is in this category for a few reasons-- it doesn't render and serves primarily to provide a line break opportunity and stop shaping/combining behavior, which will happen if used via fallback. That said we don't want to make life unnecessarily difficult in environments that don't provide fallback, so perhaps this should be added if it is commonly used.

200b is used here to break the Khanda ligatures for Ta, Na, and Ma. A special ligature is formed for these using only the base + virama, but it's possible to have base + visual virama in the same text. So to break the ligature we put in 200b before the virama.

PR #437, opened 2017-08-10
Name fixes

dougfelt:


nototools/noto_fonts.py
```diff
@@ -127,7 +127,8 @@ def script_name_for_report(script_key):
     'Thin': 100,
     'ExtraLight': 200,
     'Light': 300,
-    'DemiLight': 350,
+    'DemiLight': 350,  # used in cjk fonts
+    'SemiLight': 350,  # used in other noto fonts
```
jungshik: just  out of curiosity:    didnt' know that we generate a named instance at w=350 for non-CJK Noto.  Was this weight requested by someone?

dougfelt: Hmmm, thought SemiLight was present in some fonts, but I don't see it.  There must have been some auto-generated data that had this that the code was tripping on.  I'll track that down and remove it.

PR #439, opened 2017-08-10
Name fixes 2

dougfelt:


dougfelt:
This changes name expectations so will cause additional name lint errors until we rebuild fonts with the new pipeline.  Still need to swat sources for the fonts with special family names, too.

PR #441, opened 2017-08-16
Update family name info for phase 3.

dougfelt:
This tracks changes previously made to noto_names.py.

- all families use 'Regular' in the full name
- all families are named to accommodate all wws style variations
- since we assume all style variations, all fonts get preferred names

Pretty much all fonts will get shortened family names as a result.

The familyname_and_styles data was updated slightly to reflect decisions
on naming.  Note that it still contains some old names as well as new ones.

The new phase 3 xml data was generated with the following:
python check_familyname_and_styles.py -w | \
  xargs ./noto_names.py write -x -p 3 -f

### DIFF
--- a/nototools/data/family_name_info_p3.xml
+++ b/nototools/data/family_name_info_p3.xml
@@ -1,204 +1,215 @@
 <?xml version='1.0' encoding='utf8'?>
-<family_name_data date="2016-12-14">
-  <info family="arimo-lgc" />
-  <info family="cousine-lgc" />
-  <info family="emoji-zsye" />
-  <info family="emoji-zsye-color" />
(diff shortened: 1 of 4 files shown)


Kanjilearner817:
I'm surprised that "Noto Music" wasn't among the additions, considering that the Phase III CMAPs have code points reserved for a music font.

dougfelt:
@kanjilearner817 thanks, you're right.  It doesn't seem to be on our internal tracking, which is what the family name data was based off of.  @marekjez86, I suspect Music was originally planned for symbols but we broke it out because of different constraints on lineheight etc.  Can you check?

PR #442, opened 2017-09-06
Added Indic Number Forms to Assigned Primaries

Kanjilearner817:
I noticed that the block "Common Indic Number Forms" was absent from script assignments, so I went ahead, added the block to Assigned Primaries, and assigned it to Devanagari. Is this acceptable?


dougfelt:
This is not necessary, these are already assigned to numerous Indic scripts including Devanagari via script extensions data.

PR #452, opened 2017-09-25
Get noto update

dougfelt:
This is a collection of changes triggered by the push to update the noto site with new phase3 fonts.

The new language support required some new sample text, which lead to updating the UDHR and CLDR data.  This in turn triggered some changes in the tools that process this data and some patches.  This also adds a few new sample files for languages these updates still didn't cover.  The split of Zsym support across Symbols and Symbols2 required additional changes to how sample files were selected.  The addition of an additional 'stretch' axis (condensed, semi-condensed...) also drove some changes.


nototools/generate_website_2_data.py
```diff
@@ -435,38 +455,37 @@ def get_sample_infos(lang_scr):
...
-  if lang != 'und':
+  def add_samples(lang_scr, typ):
+    for src_key in get_sample_names_for_lang_scr_typ(lang_scr, typ):
+      sample_text = get_sample_from_sample_file(src_key)
+      if sample_text is not None:
+        attr = get_attribution(src_key)
+        sample_infos.append((sample_text, attr, src_key))
+
+  def add_exemplars(lang_scr):
     exemplar, src_key = cldr_data.get_exemplar_and_source(lang_scr)
```
jungshik: Just a thought:  would it be better to include auxil. exemplar as well?  Well, for some languages, aux. exemplar chars have ZWJ/ZWNJ and can be messy so that it needs some filtering for display. Perhaps, not worth the trouble.

dougfelt:
Yeah, sorry I'd found a few more small changes that I pushed into this batch.

nototools/generate_website_2_data.py
dougfelt: Can you point me at a file with some?

dougfelt:
Thanks a lot.  This should be the last big batch of reviews for awhile...

nototools/generate_website_2_data.py
jungshik: You mean ZWJ/ZWNJ?  For instance, Hindi has ZWJ/ZWNJ in AuxExemplarCharacters. (most other Indic languages do, too.  Bengali has a few more 'real' characters besides ZWJ/ZWNJ in AuxExemplar).

jungshik: common/main/hi.xml has this:
```
  <exemplarCharacters type="auxiliary">[\u200C\u200D]</exemplarCharacters>
```

jungshik: common/main/bn.xml has 
```
                <exemplarCharacters type="auxiliary">[\u200C\u200D ৲ ৳ ৴ ৵ ৶ ৷ ৸ ৹ ৰ ৱ]</exemplarCharacters>
```

jungshik: Not saying this has to be done now, but only including gC={L,M*,N} should work.

PR #454, opened 2017-09-25
Yi cmap

dougfelt:
Combines an update from MTI adding some chars to Coptic and designating more chars as acceptable for fallback, and an update from us requiring two new characters for Yi.


dougfelt:
Urg.  That's unhappy.  The Coptic EPACT numbers are, as I said, in SYM2 and Coptic (see the noto_cmap_phase3.txt file) but the comparison tool is listing Arab and not SYM2.  I need to figure out why.

dougfelt:
Um, oh, ok.  Misread it myself.  This shows a lot of information, and the scripts in parentheses in this case are the scripts listed in for this character in the unicode data (ScriptExtensions).  There's a different list of scripts in parentheses (this list starts with 'in') that lists the fonts in which the character appears, also, unfortunately, by script code.  This list is only shown when a character is removed (to verify that fonts still support the character), not when a character is added.  For no good reason that I can see, I should add.  Anyway in this case the font list was not shown but the script list was shown, and they look very much like the same list, but they're not.

dougfelt:
Updated the comparison script to make the two lists look a little more different.

PR #455, opened 2017-10-04
gsub_diff: Compare LookupTypes 2, 3, 4, 7

m4rc1e:
Currently [gsub_diff.GsubDiffFinder](https://github.com/googlei18n/nototools/blob/master/nototools/gsub_diff.py) can only parse and compare LookupType 1 Single subs. We cannot compare more complicated lookups. I'm in the middle of checking a collection of families and discovered this limitation.

This pr will add support for the aforementioned lookup types.

My tests use some of the examples mentioned in the [.fea spec](http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#5.a). I've also added tests for the lookup types this pr does not support; ensuring they don't get parsed incorrectly.

I decided against adding support for LookupTypes 5 and 6 because parsing OT features using regex and .ttx files is rather messy. I'd prefer notodiff and its corresponding modules to get refactored once fontTools implements https://github.com/fonttools/fonttools/issues/468 fully. It can currently 'build' features for not 'unbuild'

I apologise for the size of this pr, its pretty much a complete rewrite.

If you accept this pr, I'd like to add support for finding missing features as well. this pr will only find new/missing rules which is the same behaviour as before.

### DIFF
--- a/nototools/gsub_diff.py
+++ b/nototools/gsub_diff.py
@@ -44,41 +44,144 @@ class GsubDiffFinder(object):
 
     def find_gsub_diffs(self):
         """Report differences in substitution rules."""
-
-        rules_a = self._get_gsub_rules(self.text_a, self.file_a)
-        rules_b = self._get_gsub_rules(self.text_b, self.file_b)
-
-        diffs = []
-        report = ['']  # first line replaced by difference count
-        for rule in rules_a:
-            if rule not in rules_b:
-                diffs.append(('-',) + rule)
-        for rule in rules_b:
-            if rule not in rules_a:
-                diffs.append(('+',) + rule)
-        # ('+', 'smcp', 'Q', 'Q.sc')
+        new = [self._format_rule(r, '+') for r in self.find_new_rules()]
+        missing = [self._format_rule(r, '-') for r in self.find_missing_rules()]
+        diffs = missing + new
+        # ('+', 'smcp', 'Q', 'by', Q.sc')
         # Sort order:
         # 1. Feature tag
         # 2. Glyph name before substitution
         # 3. Glyph name after substitution
-        diffs.sort(key=lambda t:(t[1], t[2], t[3]))
+        diffs.sort(key=lambda t:(t[1], t[2], t[4]))
         report = ['%d differences in GSUB rules' % len(diffs)]
         report.extend(' '.join(diff) for diff in diffs)
         return '\n'.join(report[:self.output_lines + 1])
 
-    def _get_gsub_rules(self, text, filename):
-        """Get substitution rules in this ttxn output."""
+    def find_new_rules(self):
+        rules_a = self._get_gsub_rules(self.text_a, self.file_a)
+        rules_b = self._get_gsub_rules(self.text_b, self.file_b)
+        return [r for r in rules_b if r not in rules_a]
 
-        feature_name_rx = r'feature (\w+) {'
-        contents_rx = r'feature %s {(.*?)} %s;'
-        rule_rx = r'sub ([\w.]+) by ([\w.]+);'
+    def find_missing_rules(self):
+        rules_a = self._get_gsub_rules(self.text_a, self.file_a)
+        rules_b = self._get_gsub_rules(self.text_b, self.file_b)
+        return [r for r in rules_a if r not in rules_b]
 
-        rules = set()
-        for name in re.findall(feature_name_rx, text):
-            contents = re.findall(contents_rx % (name, name), text, re.S)
-            assert len(contents) == 1, 'Multiple %s features in %s' % (
-                name, filename)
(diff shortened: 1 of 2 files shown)


dougfelt:
@anthrotype Can you look at this?

nototools/gsub_diff.py
```diff
@@ -44,41 +44,144 @@ def __init__(self, file_a, file_b, output_lines=20):
...
-        diffs = []
-        report = ['']  # first line replaced by difference count
-        for rule in rules_a:
-            if rule not in rules_b:
-                diffs.append(('-',) + rule)
-        for rule in rules_b:
-            if rule not in rules_a:
-                diffs.append(('+',) + rule)
-        # ('+', 'smcp', 'Q', 'Q.sc')
+        new = [self._format_rule(r, '+') for r in self.find_new_rules()]
```
anthrotype: I see that refactoring those two for loops into new `find_new_rules` and `find_missing_rules` methods make the code more readable, but you are calling the same `_get_gsub_rules` method twice now, which may be an expensive operation. Perhaps you could compute `rules_a` and `rules_b` once and pass them as arguments to the previous methods.. but then again the methods would just become a label for a list comprehension (readable enough from POV). Well, up to you.

nototools/gsub_diff.py
```diff
@@ -44,41 +44,144 @@ def __init__(self, file_a, file_b, output_lines=20):
...
+    def _get_gsub_features(self, text):
+        features = set()
         feature_name_rx = r'feature (\w+) {'
-        contents_rx = r'feature %s {(.*?)} %s;'
-        rule_rx = r'sub ([\w.]+) by ([\w.]+);'
 
-        rules = set()
         for name in re.findall(feature_name_rx, text):
-            contents = re.findall(contents_rx % (name, name), text, re.S)
-            assert len(contents) == 1, 'Multiple %s features in %s' % (
```
anthrotype: the previous code would assert that only one feature block with a given name exists (which might not be true, at least for handwritten, non-ttxn-generated, feature files, since the spec allows multiple feature blocks with same name).
In your `_get_feature_content` below you have removed the assert and simply returning the result of `re.findall`, which may contain more than one match. But you are only retaining the first one, since you do `contents[0]` later on.
Was this intentional?
If we only support reading lookups from the first feature block that matches, then we need to assert that there's only one. Probably, it would be better to collect lookup rules for all the feature blocks with that name.

nototools/gsub_diff.py
```diff
@@ -44,41 +44,144 @@ def __init__(self, file_a, file_b, output_lines=20):
...
+        s = [sign]
+        for item in rule:
+            if not isinstance(item, str):
+                for sub_item in item:
+                    s.append(sub_item)
+            else:
+                s.append(item)
+        return s
+
+    def _gsub_rule_group_to_string(self, seq):
```
anthrotype: I find defining a method a bit overkill for such a trivial operation as `seq[1:-1]`, but I don't mind.

m4rc1e:
@anthrotype Thanks for this. We are working on a design proposal for modifying the diff objects. I'm aware of the repetition. We're probably going to split the OT parsing, diffing and reporting into separate objects. 

I'd keep this open for now but expect more changes, to these diff modules.

nototools/gsub_diff.py
```diff
@@ -44,41 +44,144 @@ def __init__(self, file_a, file_b, output_lines=20):
...
+            if left.startswith('[') and left.endswith(']'):
+                left = self._gsub_rule_group_to_string(left)
+
+            if right.startswith('[') and right.endswith(']'):
+                right = self._gsub_rule_group_to_string(right)
+
+            if op == 'by': # parse LookupType 1, 2, 4
+                parsed.append((feature, left.split(), op, right.split()))
+            elif op == 'from': # parse LookupType 3
+                for glyph in right.split(): # 'a.alt a.sc' -> ['a.alt', 'a.sc']
```
anthrotype: comment is redundant

nototools/gsub_diff.py
```diff
@@ -44,41 +44,144 @@ def __init__(self, file_a, file_b, output_lines=20):
...
+                    parsed.append((feature, left.split(), op, [glyph]))
```
anthrotype: `left.split()` shouldn't be necessary since in lookuptype 3 alternate substitutions the format is `substitute <glyph> from <glyphclass>;` so the left slot can contain only one glyph. A `left.strip()` would do.

anthrotype:
@m4rc1e status?

m4rc1e:
@anthrotype Unfortunately I've let this rot. I moved away from using ttxn. I don't know whether I should just spend an hour to finish this off?

I'm still looking into otlLibunbuild ;-)

PR #456, opened 2017-10-05
gpos_diff: Fixed parsing mark positions

m4rc1e:
In the previous implementation, a mark positioning rule for a single glyph wouldn't get parsed.

By adding ? to the regular expression in the method find_positioning_diffs, it can now parse both single glyphs, classes and groups.

The new test, *test_mark_positioning_diffs_simple* will test this functionality. I've added extra tests to ensure positioning diffs are being reported correctly as well.

I have renamed the existing kerning tests to avoid confusion.

If this pr gets approved, I will cleanup the helper methods in the tests _expect_kerning_diffs and _expect_mark_positioning_diffs, there's a lot of overlap.

### DIFF
--- a/nototools/gpos_diff.py
+++ b/nototools/gpos_diff.py
@@ -139,7 +139,7 @@ class GposDiffFinder:
         unmatched = {}
         mismatched = {}
         rx = re.compile('pos %s \[([\w\d\s@_.]+)\]\s+<anchor (-?\d+) (-?\d+)> '
-                        'mark (@[\w\d_.]+);' % mark_type)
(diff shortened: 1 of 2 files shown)


dougfelt:
@anthrotype This one too.

anthrotype:
For some reasons, ttxn does not add the "@" in front of mark classes when they only contain a single glyph.

https://github.com/adobe-type-tools/afdko/blob/1e5ab6dbb319b616ba1d4266f945ff2ab90db0d2/FDK/Tools/SharedData/FDKScripts/ttxn.py#L1608

By making the symbol optional with `@?` the pattern now matches.

PR #461, opened 2017-10-26
Website update 2

dougfelt:
Numerous small changes for website update.

### DIFF
--- a/nototools/noto_fonts.py
+++ b/nototools/noto_fonts.py
@@ -419,6 +419,9 @@ def wws_family_id_to_name_parts(wws_id):
       parts.append('kr')
   elif script == 'sym2':
     parts.append('Symbols2')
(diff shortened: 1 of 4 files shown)


nototools/noto_fonts.py
```diff
@@ -419,6 +419,9 @@ def wws_family_id_to_name_parts(wws_id):
       parts.append('kr')
   elif script == 'sym2':
     parts.append('Symbols2')
+  elif script == 'phag':
+    # allow hyphenated name in name table
+    pargs.append('Phags-pa')
```
jungshik: pargs?  Isn't it |parts|?

dougfelt: Um... this must not have executed in the tools I ran.  Hmmm.  I'll leave this in and fix it instead of removing it entirely.

PR #468, opened 2018-03-03
updated handling of name/designer

marekjez86:
Wanted to handle currently existing name/designer strings


marekjez86:
@dougfelt : I'm not sure that this is the right way of doing changes and requesting reviews. If yes, please review this small change. If not, please educate me what needs to happen.

marekjez86:
@jungshik : I'm not sure if dougfelt is around. Could you look at it?

PR #473, opened 2018-08-23
Update environment setup for emoji tooling

dougfelt:


anthrotype:
@dougfelt If I grep for "scour" I can't find anything in the repo. Where is this `scour` tool being used and how?

anthrotype:
oh, I see it's used in noto-emoji repo:
https://github.com/googlefonts/noto-emoji/blob/master/scour_svg.sh#L46

but then it's technically not a requirement of nototools per se. I'll see if I can move this where it belongs

PR #478, opened 2019-02-19
Fixes and updates to noto lint

punchcutter:
Added newer Unicode scripts to Noto lint.

Updated Sans Balinese and Sans Ahom to Serif Balinese and Serif Ahom to match the name changes in noto-source and noto-fonts. 

Updated fonttools requirements to fix noto_lint failure when calling subset. The fonttools Subsetter used to have a `glyphs_all` property which was renamed to `glyphs_retained` in v36.0.0. That broke noto_lint so updated the fonttools requirements and the noto_lint use of `glyphs_retained`. 

Updated some print statements because currently nototools is nowhere near Python 3 compatible. Any reason why it shouldn't be?


punchcutter:
@dougfelt @anthrotype Can someone merge these pull requests? There are 4 from 2017 that never got merged and then this one.

anthrotype:
yeah sorry about that. It appears that nototools is in need of a maintainer.

PR #484, opened 2019-04-22
Cmap correction for Ahom

Kanjilearner817:
Character count should be 63 (was 62).


Kanjilearner817:
I've just looked up that U+1171A was added in Version 11.0. Pursue addition policy for other scripts already on list?

punchcutter:
This PR can be closed because the change was included in https://github.com/googlefonts/nototools/pull/497

Issue #472, opened 2018-06-25
Please consider to switch to python 3.x


epico:
Currently we are switching to python 3.x in Fedora, please consider to switch nototools to use python 3.x.

Thanks in advance!

PureTryOut:
More and more distributions are dropping Python 2 since it's going to be end of life soon. Please consider dropping Python 2 support soon and switching to Python 3!

ILMostro:
third-party tools still need to be ported; e.g. spiro and fontcrunch.

PR #488, opened 2019-07-16
Make it work both with Python2 and Python3

mike-fabian:
With this change (and some changes I did to the python code in https://github.com/googlei18n/noto-emoji.git I was able to build the "NotoColorEmoji.ttf" font using Python3.

### DIFF
--- a/nototools/collect_cldr_punct.py
+++ b/nototools/collect_cldr_punct.py
@@ -99,7 +99,7 @@ def _collect_script_to_punct(files):
       script = cldr_data.get_likely_script(filename)
       if script == 'Zzzz':
         if filename != 'root':
-          print >> sys.stderr, 'no script for %s' % filename
+          sys.stderr.write('no script for %s\n' % filename)
       else:
         script_to_punct[script] |= punct
 
@@ -135,13 +135,13 @@ def script_to_punct():
 
 
 def _write_script_to_punct(script_to_punct):
-  print 'SCRIPT_TO_PUNCT = {'
+  print('SCRIPT_TO_PUNCT = {')
   for script in sorted(script_to_punct):
     chars = script_to_punct[script]
     int_chars = [ord(cp) for cp in chars]
-    print '  # %s' % ('|'.join(sorted(chars)))
-    print "  '%s': '%s'," % (script, tool_utils.write_int_ranges(int_chars))
-  print '}'
+    print('  # %s' % ('|'.join(sorted(chars))))
+    print("  '%s': '%s'," % (script, tool_utils.write_int_ranges(int_chars)))
+  print('}')
 
 
 def main():
--- a/nototools/generate_sample_text_html.py
+++ b/nototools/generate_sample_text_html.py
@@ -71,9 +71,9 @@ def generate_table(filename):
         cols.append(sample_type)
         cols.append(sample_text)
         f.write('<td>'.join(cols) + '\n')
-      f.write('<tr><td colspan=4>&nbsp;')
-    f.write('</table>')
-    f.write(_HTML_FOOTER)
+      f.write('<tr><td colspan=4>&nbsp;\n')
+    f.write('</table>\n')
+    f.write(_HTML_FOOTER + '\n')
 
 
 def _get_script_to_samples():
--- a/nototools/generate_sample_text_html.py
+++ b/nototools/generate_sample_text_html.py
@@ -44,13 +44,13 @@ _HTML_FOOTER = """</body>
 def generate_table(filename):
   with codecs.open(filename, 'w', 'utf-8') as f:
     script_to_samples = _get_script_to_samples()
-    print >> f, _HTML_HEADER
-    print >> f, '<table>'
-    print >> f, '<tr><th>Script<br/>BCP<th>name<th>type<th>text'
+    f.write(_HTML_HEADER)
+    f.write('<table>\n')
+    f.write('<tr><th>Script<br/>BCP<th>name<th>type<th>text\n')
 
     for script, samples in sorted(script_to_samples.iteritems()):
       script_en = cldr_data.get_english_script_name(script)
-      print >> f, '<tr><th colspan=4>%s' % script_en
+      f.write('<tr><th colspan=4>%s\n' % script_en)
       for bcp, sample_type, sample_text in samples:
         try:
           lsrv = cldr_data.loc_tag_to_lsrv(bcp)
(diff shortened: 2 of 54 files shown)


nototools/generate_sample_text_html.py
```diff
@@ -62,18 +62,18 @@ def generate_table(filename):
...
         cols.append(bcp_no_script)
         cols.append(bcp_en)
         cols.append(sample_type)
         cols.append(sample_text)
-        print >> f, '<td>'.join(cols)
-      print >> f, '<tr><td colspan=4>&nbsp;'
-    print >> f, '</table>'
-    print >> f, _HTML_FOOTER
+        f.write('<td>'.join(cols) + '\n')
+      f.write('<tr><td colspan=4>&nbsp;')
```
dougfelt: why no '\n' added in these lines?

nototools/noto_cmap_reqs.py
```diff
@@ -3285,10 +3285,10 @@ def _dump_primaries():
   for block in unicode_data.block_names():
     block_range = unicode_data.block_range(block)
     primary_script = _primary_script_for_block(block)
-    print '%13s %6s %s' % (
+    print('%13s %6s %s\n' % (
```
dougfelt: It seems the '\n' is not wanted here?  Print already issues a newline, right?

nototools/sync_repos.py
```diff
@@ -84,17 +84,17 @@ def noto_checkout(
       failed_tags.append('%s: %s' % (r, t))
 
   if failed_tags:
-    print >> sys.stderr, 'failed to find:\n  %s' % '\n  '.join(failed_tags)
+    sys.stderr.write('failed to find:\n  %s' % '\n  \n'.join(failed_tags))
```
dougfelt: The added newline is in the wrong place.

nototools/unicode_data.py
```diff
@@ -784,7 +784,7 @@ def _read_emoji_data(lines):
       continue
 
     # discourage lots of redundant copies of seq_type
```
dougfelt: delete comment, no longer applies?

nototools/generate_sample_text_html.py
dougfelt: meant to make that a review comment

mike-fabian: Oh, you are right! adding these '\n' ...

nototools/noto_cmap_reqs.py
mike-fabian: You  are right! Sorry!

nototools/sync_repos.py
mike-fabian: Yes!

nototools/unicode_data.py
mike-fabian: Yes, no longer applies when  the intern() is gone!

PureTryOut:
So this will close #472 when merged then.

mike-fabian:
Do I need to do something about the “Need a CLA for one or more commit authors”?

The commits have mfabian@redhat.com as the author e-mail, should I change that to my Google mail maiku.fabian@gmail.com because I used that to sign the CLA? I tried to add mfabian@redhat.com as an additional address to my google account, but that didn’t work because it is used in my work account and apparently one cannot add the same address to another google account.

dougfelt:
Hmmm, I'm not sure about workarounds for this.  If you can change the
commit address that would likely work.

mike-fabian:
> Hmmm, I'm not sure about workarounds for this. If you can change the commit address that would likely work.

I have changed the commit address.

RoelN:
I can't get `pip3 install -r requirements.txt` to install properly. I'm on macOS High Sierra. It errors out on:

  ```
Building wheels for collected packages: pyclipper
  Building wheel for pyclipper (setup.py) ... error
  [...]
    2 warnings and 1 error generated.
    error: command 'clang' failed with exit status 1
    ----------------------------------------
    ERROR: Failed building wheel for pyclipper
    Running setup.py clean for pyclipper
    Failed to build pyclipper
  [...]
  ```

and then:

```
ERROR: ufo2ft 2.9.1 has requirement booleanOperations>=0.8.2, but you'll have booleanoperations 0.7.0 which is incompatible.
ERROR: ufo2ft 2.9.1 has requirement defcon>=0.6.0, but you'll have defcon 0.3.1 which is incompatible.
ERROR: mutatormath 2.1.2 has requirement defcon>=0.3.5, but you'll have defcon 0.3.1 which is incompatible.
ERROR: fontmake 1.10.1 has requirement booleanOperations>=0.8.2, but you'll have booleanoperations 0.7.0 which is incompatible.
ERROR: fontmake 1.10.1 has requirement defcon[lxml]>=0.6.0, but you'll have defcon 0.3.1 which is incompatible.
```

I eventually managed to get al requirements in order by manual installs through pip3 and brew. I was able to run:

`sudo python3 setup.py develop`

So while it's working now (a Noto Emoji build is happily purring in the background), I'm not sure if I'm doing something wrong, the requirements.txt need to be modified or something else is wrong. Any advice?

RoelN:
If nobody else is running into problems, bite the bullet and go full python 3? @dougfelt

RoelN:
https://github.com/googlefonts/noto-emoji/pull/267 is waiting for this PR. Can we merge this?

nototools/unicode_data.py
```diff
@@ -783,8 +783,7 @@ def _read_emoji_data(lines):
     if m.group(2):
       continue
 
-    # discourage lots of redundant copies of seq_type
-    seq_type = intern(m.group(3).strip().encode('ascii'))
```
anthrotype: FWIW the old `intern` builtin was moved to `sys.intern`: https://docs.python.org/3/library/sys.html?highlight=sys.intern#sys.intern

nototools/collect_cldr_punct.py
```diff
@@ -99,7 +99,7 @@ def _collect_script_to_punct(files):
       script = cldr_data.get_likely_script(filename)
       if script == 'Zzzz':
         if filename != 'root':
-          print >> sys.stderr, 'no script for %s' % filename
+          sys.stderr.write('no script for %s\n' % filename)
```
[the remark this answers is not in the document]
plugwash: sys.stderr.write(foo) works in both python2 and python3, print(foo, file=sys.stderr) only works in python3.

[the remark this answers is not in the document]
anthrotype: You need to `from __future__ import print_function`.
If we haven't done that yet, we should make sure to add this as well as the other two canonical future imports (`absolute_import`, `division`) to all modules.

PR #492, opened 2019-09-23
Complete se-Latn_udhr.txt

AlsoScratch:
Translation from Omniglot.


AlsoScratch:
@googlebot I signed it!

PR #502, opened 2020-01-09
Python3 support for spiro

RaspberryPiFan:
This makes spiro comaptible to python3, but it should still work with python2

PR #503, opened 2020-01-09
Remove python2

RaspberryPiFan:
This removes python2 due the stopped support.

> DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.


RaspberryPiFan:
This PR needs #502 to work.

Issue #501, opened 2019-12-15
third_party/spiro looks to be not compatible with python3


pacho2:
I am trying latest git with python3 support, but, at install time, I see lots of errors like this that look to show that third_party/spiro is not yet ready for python3
Installing notocoverage script to /var/tmp/portage/dev-python/nototools-20191107/image/_python3.7/usr/lib/python-exec/python3.7
Installing notodiff script to /var/tmp/portage/dev-python/nototools-20191107/image/_python3.7/usr/lib/python-exec/python3.7
  File "/usr/lib/python3.7/site-packages/third_party/spiro/x3/pyrex/main.py", line 4
    print cmd, what, arg
            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(cmd, what, arg)?

  File "/usr/lib/python3.7/site-packages/third_party/spiro/x3/pyrex/beztest.py", line 5
    print cmd, what, arg
            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(cmd, what, arg)?

  File "/usr/lib/python3.7/site-packages/third_party/spiro/font/replace_class.py", line 14
    print fn, table[clas]
           ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(fn, table[clas])?

  File "/usr/lib/python3.7/site-packages/third_party/spiro/font/mkblends.py", line 14
    print cmd
            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(cmd)?

Sorry: TabError: inconsistent use of tabs and spaces in indentation (tocubic.py, line 63)
  File "/usr/lib/python3.7/site-packages/third_party/spiro/curves/polymat.py", line 57
    print '%6.1f' % m[j][i],
                ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('%6.1f' % m[j][i], end=" ")?

  File "/usr/lib/python3.7/site-packages/third_party/spiro/curves/polymat-bad.py", line 53
    print '%6.1f' % m[j][i],
                ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('%6.1f' % m[j][i], end=" ")?

  File "/usr/lib/python3.7/site-packages/third_party/spiro/curves/poly3.py", line 85
    print 306 + 300 * x, 400 + 300 * y, cmd
            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(306 + 300 * x, 400 + 300 * y, cmd)?

[...]

PR #506, opened 2020-01-21
Partial python3 support

### LINKED ISSUE #472: Please consider to switch to python 3.x

penguinolog:
Fix #472  #501 

Fixed:
* print usage
* file operations (`file` is not supported anymore)
* `from math import *`
* mix of tabs and spaces
* python 2.4- code has been removed. Check shown, that it is not used.
* `md5` import for python 3: use `hashlib.md5`
* raising string (in python3 `Exception` should be used)
* using backtick instead of `repr`
* use `from __future__ import division` and `//` for integer division


penguinolog:
3 CI builds was failed due to some error in PIP (tests was not started)

PR #509, opened 2020-01-22
 Normalize code style. Fix #508 

### LINKED ISSUE #508: Consider enforcement of code style
`nototools` is a package with history and contains multiple code styles, including proof of concept versions. This situation causes difficulties for visual code analisys.

Idea:
1. 120 symbols per line
2. Use `isort` with one import per line
3. Use `black` for automatic formatting.

I expect, that migration commit will be huge as hell ant will cause merge conflicts for not merged PRs, but next work on tooling will be much easier.

penguinolog:
1. Enforce `black` code style
2. Apply `isort` to code and remove unused imports
3. Fix flake8 and flake8-bugbear for nototools

* unused variable -> comment/remove
* unused loop control variable -> use `_` or prefix `_`
* bare except -> use `except Exception`
* `not var in Container` -> `var not in Container`
* `assert False` -> `raise AssertionError`
* str() object is str() object -> use `==`
* use `isinstance` for type checking

* undefined variable -> make quiet and add TODO
* mutable defaults -> make quiet and add TODO


penguinolog:
Note: first 2 commits at this PR are automatic generated using corresponding tools, while last one is result of hand work.
Technically `flake8` validation may be enabled for `nototools` package after this PR.

marekjez86:
@penguinolog : you've done a great job here. Thank you. What stops me from approving it: this was developed by the Google engineers. The convention used for Python is to limit the line length to 80 characters (with very few exceptions)... After your changes some of the lines exceed that. For example, "__version_info_re = re.compile(r'GOOG;noto-(?:fonts(?:-alpha)?|source):(\d{4})(\d{2})(\d{2}):([0-9a-f]{12})')_" , if I counted correctly, has 109 characters. If I approve it, this will go against the grain of the original developer style. Is there a way to limit the line length to 80 chars?

penguinolog:
Yes, I can do it in semiautomatic mode.

penguinolog:
Enforcement of 79 symbols/line (standard PEP8) looks not the bes idea for this code:
1. huge amount of nested if's and cycles causes serious offset and with such limit we have worse readability due to line breaks
2. a lot of really long string constants and comments (and some of them is not allowed to re-format)

Can we review specific rules for repo?

marekjez86:
@penguinolog : please do

penguinolog:
Switched to 79 symbols/line, but some parts of code now looks ugly.
And due to nesting amount + constants/method names in some places code not fit even in 100 symbols.

penguinolog:
Fixed as much as was possible on this stage, but feels like code requires serious rework.
And I expect bunch of errors during real python3 usage: not all places are easy to find statically.

anthrotype:
May I ask what problem are you trying to solve here? You seem to be making stylistic changes for the sake of it. Are you even using these tools?

penguinolog:
I'm using it for local building of noto-emoji (on gentoo) and I want to drop python2.7 from system.
Unfortunatelly, porting this code to python3 is not so simple task: test code coverage is poor and static analysis ignores some errors. I had to spend huge amount of time just for reading code and fixing "trivial" incompatibilities like trying to join iterators as lists without casting & etc, and it was not easy.

anthrotype:
Thank you. I don't see how running an auto-formatter would make it easier to port the code to python3. I agree  that test coverage is not ideal, but making sweeping changes like these might increase the chances of introducing regressions. Reviewing those changes also costs time. I agree that using a normalized style and automatically enforcing that is a good idea for *new* projects. Less so for existing ones that haven't consistently followed those rules. Besides, the original authors and maintainers if nototools have moved to different projects. As for me, I am also busy with other things. But I definitely welcome any patches that improve python3 support (I thought it was done already. If that's not the case, please open specific issues, if you haven't already done so).

penguinolog:
Auto-formatter helped to make code human-readable, because it was really difficult to read.
Technically I can try to separate found issues, like unresolved references and extra TypeErrors, but re-writing it again will be not easy

PR #512, opened 2020-04-28
Find hb-shape using the user's PATH environment variable

m4rc1e:
Current implementation assumes the path to `hb-shape` is located at, `/Users/user /harfbuzz/util/hb-shape`. This assumption won't be correct for every user so a better approach is to first check whether the user has `hb-shape` as a symlink or if it exists in a `bin` dir. If it doesn't exist, then we can use the existing path for backwards compatibility reasons.

I realise that python3 has `shutil.which` but I've opted not to use this since Roboto v1/2 is still py2. Using `distutils.spawn.find_executable` works on both py2 and py3.

### DIFF
--- a/nototools/render.py
+++ b/nototools/render.py
@@ -131,15 +131,13 @@ def get_glyph_vertical_extents(glyph_id, font_file_name):
     return get_glyph_cleaned_extents(ttglyph, glyf_set)
 
 
-# FIXME: figure out how to make this configurable
-HARFBUZZ_DIR = os.getenv('HOME') + os.sep + 'harfbuzz'
-HB_SHAPE_PATH = HARFBUZZ_DIR + os.sep + 'util' + os.sep + 'hb-shape'
+HB_SHAPE_DIR = os.path.join(os.getenv('HOME'), 'harfbuzz', 'util')
 
 
 def run_harfbuzz_on_text(text, font_file_name, language, extra_parameters=None):
     """Runs HarfBuzz on input text and return JSON shaping information."""
     hb_parameters = [
-        HB_SHAPE_PATH,
+        "hb-shape",
         '--output-format=json',
         '--no-glyph-names',  # Some fonts have empty glyph names
         '--font-file=%s' % font_file_name]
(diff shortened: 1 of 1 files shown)


m4rc1e:
Auch. The tests have been broken since 3359af33c65d4eb231587f7436ba3f45a5037de8. I'll have a dig around soonish.

anthrotype:
I wonder what's the reason for the hardcoded path. Can't we just let subprocess figure out automatically where `hb-shape` is? If we pass just the name of the executable, I believe it will call whatever is first in the user's $PATH, no?

m4rc1e:
> I wonder what's the reason for the hardcoded path.

It currently has a TODO note so perhaps they simply didn't have enough time. I've kept the existing path just incase if we can't find `hb-shape`, in case my assumption is wrong.

> Can't we just let subprocess figure out automatically where hb-shape is?

subprocess will raise a `FileNotFoundError` if `hb-shape` doesn't exist. We could wrap it in a try/accept but I feel this is a cleaner approach.

anthrotype:
Personally, I would have caught the the FileNotFoundError exception (I believe for python2-compat you'd have to catch IOError and check the exception's `errno` attribute != errno.ENOENT).
One doesn't need to import some obscure distutils function (used mostly for internal use by distutils) to use the subprocess API.
But I don't mind if you prefer to fix it this way.

m4rc1e:
Forgot to add that if `hb-shape` does exist and we check its existence using any subprocess function, we'll get terminal output e.g

```
>>> subprocess.call(['hb-shape'])
Usage: hb-shape [OPTION...] [FONT-FILE] [TEXT]
```

We can silence it by redirecting the output to DEVNULL but this is also messy.

```
import os
import subprocess

FNULL = open(os.devnull, 'w')
subprocess.call(['hb-shape'], stdout=FNULL, stderr=subprocess.STDOUT)
```

Tbh, we're going to overwrite Roboto v1 soon so it may make sense to just use `shutil.which` and forget about py2 compatibility.

m4rc1e:
After speaking to Cosimo, we decided not to use distutils or which. Our new approach is duplicate the users PATH env, append the hardcoded hb-shape path to it and then exec subprocess using just `hb-shape` and forcing it to use the new environment variables.

PR #514, opened 2020-04-30
Fix unicode_data_test

m4rc1e:
travis fails recent pull requests because the repo's unicode data had recently been upgraded v13.0 but the checks were never updated.


m4rc1e:
Since version 13.0, there's only 64 code points in unallocated ranges (48 code points at 0870..089F and 16 code points at 2FE0..2FEF) so I've added 0x0870.

anthrotype:
thanks

Issue #530, opened 2021-01-26
installation fails on MacOS


Shachlan:
When trying to execute the `pip install -r requirements.txt` step in the readme, I get these 2 exceptions:
1. `ERROR: Failed building wheel for Pillow`:
```
Building wheels for collected packages: Pillow, pyclipper
  Building wheel for Pillow (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/local/opt/python@3.9/bin/python3.9 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/6r/4ynwf0ld5_lb39f7qxvvv07w0000gn/T/pip-install-uzmxa25z/pillow_8fbcb8609915481597a44a4d1b767a9d/setup.py'"'"'; __file__='"'"'/private/var/folders/6r/4ynwf0ld5_lb39f7qxvvv07w0000gn/T/pip-install-uzmxa25z/pillow_8fbcb8609915481597a44a4d1b767a9d/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/6r/4ynwf0ld5_lb39f7qxvvv07w0000gn/T/pip-wheel-d_8u7n_9
       cwd: /private/var/folders/6r/4ynwf0ld5_lb39f7qxvvv07w0000gn/T/pip-install-uzmxa25z/pillow_8fbcb8609915481597a44a4d1b767a9d/
  Complete output (172 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.macosx-10.15-x86_64-3.9
  creating build/lib.macosx-10.15-x86_64-3.9/PIL
  copying src/PIL/MpoImagePlugin.py -> build/lib.macosx-10.15-x86_64-3.9/PIL
[...]
  copying src/PIL/FtexImagePlugin.py -> build/lib.macosx-10.15-x86_64-3.9/PIL
  running egg_info
  writing src/Pillow.egg-info/PKG-INFO
  writing dependency_links to src/Pillow.egg-info/dependency_links.txt
  writing top-level names to src/Pillow.egg-info/top_level.txt
  reading manifest file 'src/Pillow.egg-info/SOURCES.txt'
  reading manifest template 'MANIFEST.in'
[...]

PR #536, opened 2021-04-14
Add 'b' to the prefix to compare Bytes literals

WordlessEcho:
> Bytes literals are always prefixed with 'b' or 'B'; they produce an instance of the `bytes` type instead of the `str` type. They may only contain ASCII characters; bytes with a numeric value of 128 or greater must be expressed with escapes.

https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
https://stackoverflow.com/questions/6269765/what-does-the-b-character-do-in-front-of-a-string-literal

### DIFF
--- a/nototools/ttc_utils.py
+++ b/nototools/ttc_utils.py
@@ -63,7 +63,7 @@ class TTCFile(object):
 
     def _build(self, data):
         tag, version, font_count = struct.unpack(_ttcHeader, data[:_ttcHeaderSize])
-        if tag not in ["ttcf"]:
+        if tag not in [b"ttcf"]:
             raise ValueError("not a font collection")
         if version not in [0x10000, 0x20000]:
             raise ValueError("unrecognized version %s" % version)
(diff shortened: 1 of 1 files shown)


WordlessEcho:
@googlebot I signed it!

kojiishi:
Happened to hit this issue, `ttc_utils.ttcfile_extract` does not work on Python3, and confirmed this PR fixes.

WordlessEcho:
@kojiishi Also notice this pull request #531 
Your tool [contextual-spacing](https://github.com/kojiishi/contextual-spacing) is very nice.

kojiishi:
Thanks for the feedback!

> @kojiishi Also notice this pull request #531

Looked at #530, `pip install -r requirements.txt` works for me on Ubuntu, probably Mac-only issue?

> Your tool [contextual-spacing](https://github.com/kojiishi/contextual-spacing) is very nice.

Thanks!

WordlessEcho:
> Thanks for the feedback!
> 
> > @kojiishi Also notice this pull request #531
> 
> Looked at #530, `pip install -r requirements.txt` works for me on Ubuntu, probably Mac-only issue? I don't see problems in the PR but I'm not an owner of this repo.
> 
> > Your tool [contextual-spacing](https://github.com/kojiishi/contextual-spacing) is very nice.
> 
> Thanks!

It happens on Python 3.9.

Issue #537, opened 2021-04-26
ttc_utils.ttcfile_extract ignores output_ttc_path


kojiishi:
`ttc_utils.ttcfile_extract` has following issues to run on Python3:
1. The bytes/string issue to be fixed in PR #536.
2. The current AFDKO does not use the 'FDK' directory name.
3. The current AFDKO `otc2otf` writes the output to the same directory as the input, not to the current directory.

I can think of a couple of ways to fix 2 and 3.

**1. Fix to match the current AFDKO `otc2otf` behavior**

For issue 2, the current code assumes `FDK` and `linux` in the path. Probably we can fix it by finding the program in `PATH`.

For issue 3, `otc2otf` does not have an option to specify the output directory, so we will need to move the output files to `output_ttc_path` after `otc2otf' completed.

**2. Change it to use `ttCollection` in `fonttools`**

From the git history, it looks like `ttc_utils` was originally written before `ttCollection` in `fonttools`. Now these functionalities are covered by `ttCollection`, we can rewrite `ttcfile_extract` by using `ttCollection`.

**3. Deprecate `ttcfile_extract` and `ttcfile_build`**

As far as I can find in this repo and in [AOSP noto-fonts](https://android.googlesource.com/platform/external/noto-fonts/), these two functions are used only to a) extract, b) modify OTF using `ttFont`, then c) build back to TTC.

`ttCollection` can do this operation on memory, without needing to extract to temporary files. If we can change [`patch_cjk_ttc` in `android_patches.py`](https://github.com/googlefonts/nototools/blob/main/nototools/android_patches.py#L149) and [change AOSP `subset_noto_cjk.py`](https://android-review.googlesource.com/c/platform/external/noto-fonts/+/1686105) to use `ttCollection`, we can eliminate these two functions.

`swat_license.py` uses `ttcfile_filenames` and `build_ttc`, so these two functions may need to remain, using 1 or 2.

The command line usage of `ttc_utils.py` needs to be removed too, but I think this is ok as `otc2otf` is a command line tool one can easily switch to.

@marekjez86 WDYT? I'm leaning towards 2 or 3, but feedback is appreciated.

kojiishi:
Thanks @marekjez86 !

PR #538, opened 2021-04-26
Change ttc_utils.py to use `fonttools` instead of `AFDKO`

kojiishi:
This patch changes the implementation of `ttc_utils.py` by
using `fonttools` instead of `AFDKO`.

This is the approach 2 mentioned in #537.

### DIFF
--- a/nototools/ttc_utils.py
+++ b/nototools/ttc_utils.py
@@ -22,113 +22,35 @@ from os import path
 import struct
 import subprocess
 
+from fontTools.ttLib import TTFont
 from fontTools.ttLib.tables._n_a_m_e import table__n_a_m_e as NameTable
+from fontTools.ttLib.ttCollection import TTCollection
 
 from nototools import tool_utils
 
-_ttcHeader = ">4sLL"
-_ttcHeaderSize = struct.calcsize(_ttcHeader)
-
-_sfntHeader = ">LHHHH"
-_sfntHeaderSize = struct.calcsize(_sfntHeader)
-
-_sfntHeaderEntry = ">4sLLL"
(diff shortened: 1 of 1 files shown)


nototools/ttc_utils.py
```diff
@@ -160,51 +80,58 @@ def ttc_filenames(ttc, data):
...
-        raise ValueError("can not resolve %s" % tool_path)
+def ttfont_filename(font):
+    font_fmt = ttfont_format_as_extension(font)
+    name_table = font.get("name")
+    if name_table:
+        ps_name = None
+        for r in name_table.names:
+            if (r.nameID, r.platformID, r.platEncID, r.langID) == (6, 3, 1, 0x409):
+                ps_name = r.string.decode("UTF-16BE")
+                break
```
anthrotype, line 99: fonttools name table has a getDebugName() method that gives you an English name string, you can use that to get the PostScript name

anthrotype (approved):

kojiishi:
@marekjez86

nototools/ttc_utils.py
kojiishi: Thanks for the comment. The [logic](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L103) looks a little different, are we fine to change? This logic appears in the diff because I needed to move, but it is as it was in the original code.

kojiishi: Updated the PR to use `getDebugName`.

anthrotype: effectively it does the same thing, except that it attempts to get the Macintosh English name first (if any), but that's going to be exactly the same as the Windows English string so it should not matter.

PR #540, opened 2021-04-30
Sync cmap in `patch_cjk_ttc` with AOSP

kojiishi:
This patch updates the list of code points in `patch_cjk_ttc`
to match the one in `subset_noto_cjk.py` in AOSP.


anthrotype (approved):

kojiishi:
@marekjez86 I locally confirmed this produces the same font as `subset_noto_cjk.py` in AOSP.

PR #640, opened 2022-03-30
Remove font "fixing" scripts

simoncozens:
Removes a number of binary patching scripts which give the mistaken impression that they are used as part of the Noto build process. @marekjez86 tells me that no binary patching scripts are currently used as part of generating or delivering Noto fonts.


rsheeter:
I will defer to @marekjez86 for this one, if we don't use these anymore then +1

marekjez86 (commented):
I want to keep
android_patches.py
autofix_for_release.py
fix_noto_cjk_thin.py
You can delete rest

